PR #171, opened 2012-06-26
Add tuple_cat and tests.

bryancatanzaro:
Let me know what other tests or changes you'd like.


brycelelbach:

brycelelbach:
So we're just gonna replace our tuple with a C++11 tuple or `cuda::std::tuple`, so I don't think this will make sense going forward.

PR #215, opened 2012-08-10
Add tuple_reduce, as well as min/max/sum.

bryancatanzaro:
This code was useful to me, so I thought I'd send it over.


brycelelbach:

brycelelbach:
So we're just gonna replace our tuple with a C++11 tuple or `cuda::std::tuple`, so I don't think this will make sense going forward.

PR #616, opened 2015-01-13
Mark TestTransform[Binary|Unary]CountingIterator as Known Failure with GCC 4.4

kperelygin:
Known failure when running TestTransformBinaryCountingIterator or
TestTransformUnaryCountingIterator built with GCC 4.4 due to GCC bug:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=43251


jaredhoberock:
Thanks Kyrylo -- since Thrust 1.8 is released now, this pull request should go into the master branch.

Unfortunately, I can't change the target of the request on my end. Could you switch it?

PR #617, opened 2015-01-13
Known failure when running TestTransform[Binary|Unary]CountingIterator with GCC 4.4

kperelygin:
Known failure when running TestTransform[Binary|Unary]CountingIterator with GCC 4.4 due to GCC bug:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=43251


jaredhoberock:
Thanks, that's perfect.

PR #626, opened 2015-01-27
GCC 4.3 does not do ADL correctly on malloc call

kperelygin:
The tests call back into memory.cu for get_temporary_buffer (perfect, ADL works fine) but when it’s time to allocate memory, instead of going into thrust::system::cuda::detail::malloc<thrust::system::cuda::detail::tag>, it resolves the ADL to thrust::pointer<void,my_memory_system> malloc(my_memory_system &system, std::size_t) inside memory.cu (that malloc is used for validating the malloc dispatch unit test).

Since in GCC 4.4+ it resolves to the correct call, I am marking this as a compiler bug.

### DIFF
--- a/testing/memory.cu
+++ b/testing/memory.cu
@@ -198,8 +198,9 @@ template<typename T>
 
 void TestGetTemporaryBufferDispatchExplicit()
 {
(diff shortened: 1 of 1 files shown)


jaredhoberock:
Thanks for working this out Kyrylo! Looks good.

PR #634, opened 2015-02-26
Added iterator concatenation example

sdalton1:
Simple example to concatenate two iterators.


jaredhoberock:
You might take a look at the implementation here:

https://github.com/thrust/thrust/blob/master/thrust/iterator/detail/join_iterator.h

sdalton1:
Thanks for the pointer. This does pretty much everything I need.

PR #645, opened 2015-03-11
Used UINT_MAX instead of std::numeric_limits to avoid __host__ __device_...

sdalton1:
..._ warning

The following code issues a warning when the device backend is set to OMP.

```shell
/usr/local/cuda/include/thrust/system/detail/generic/copy_if.inl(143): warning: calling a __host__ function from a __host__ __device__ function is not allowed
```

This update replaces std::numeric_limits<unsigned int>::max with UINT_MAX to silence the warning.

```c++
#include <thrust/copy.h>

int main(void)
{
    thrust::device_vector<int> A;
    thrust::device_vector<bool> stencil;

    thrust::copy_if(A.begin(), A.end(), A.begin(), thrust::identity<bool>());

    return 0;
}
```

### DIFF
--- a/thrust/system/detail/generic/copy_if.inl
+++ b/thrust/system/detail/generic/copy_if.inl
@@ -25,6 +25,7 @@
 #include <thrust/distance.h>
 #include <thrust/transform.h>
 #include <thrust/detail/internal_functional.h>
+#include <thrust/detail/integer_traits.h>
 #include <thrust/detail/temporary_array.h>
 #include <thrust/detail/type_traits.h>
 #include <thrust/scan.h>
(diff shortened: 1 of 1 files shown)


jaredhoberock:
Thanks Steven. We don't use `UINT_MAX` directly. I think it needs to be `thrust::detail::integer_traits<unsigned int>::const_max`. You need to `#include <thrust/detail/integer_traits.h>` to use that class.

sdalton1:
Works great, thanks for the pointer. I updated the implementation to use integer_traits.

jaredhoberock:
Looks good! Thanks.

PR #649, opened 2015-03-14
Added asserts to sorting_aos_vs_soa.cu to check that the results are sorted correctly

jdesjarlais:


jaredhoberock:
Thanks Jake! Looks great.

PR #654, opened 2015-04-07
Stream operator<< for device_reference

ccecka:
Implicit conversions are not considered during template argument deduction. Currently, stream output of a `device_reference<A>` where `A` is non-template (e.g. `int`, `float`, `double`, `mystruct`) functions as expected, but fails when `A` is a template (e.g. `thrust::complex<float>`, `mytemplate<int>`).

Example:

```
#include <thrust/complex.h>
#include <iostream>

int main() {
  thrust::device_vector<double> a(5);
  std::cout << a[2] << std::endl;   // Outputs '0' as expected

  thrust::device_vector<thrust::complex<double> > b(5);
  std::cout << b[2] << std::endl;   // XXX: Fails to compile despite operator<< being defined for thrust::complex<T>
}
```

This patch simply uses `static_cast` to force the conversion and mirrors `operator<<` defined for `device_ptr<T>`.

### DIFF
--- a/thrust/complex.h
+++ b/thrust/complex.h
@@ -541,8 +541,9 @@ template <typename T> __host__ __device__ complex<T> atanh(const complex<T>& z);
  *  \param os The output stream.
  *  \param z The \p complex number to output.
  */
-template<typename ValueType,class charT, class traits>
-std::basic_ostream<charT, traits>& operator<<(std::basic_ostream<charT, traits>& os, const complex<ValueType>& z);
+template<typename ValueType, typename charT, typename traits>
+std::basic_ostream<charT, traits>&
+operator<<(std::basic_ostream<charT, traits>& os, const complex<ValueType>& z);
 
 /*! Reads a \p complex number from an input stream.
  *  The recognized formats are:
@@ -555,7 +556,7 @@ std::basic_ostream<charT, traits>& operator<<(std::basic_ostream<charT, traits>&
  *  \param is The input stream.
  *  \param z The \p complex number to set.
  */
-template<typename ValueType, typename charT, class traits>
+template<typename ValueType, typename charT, typename traits>
 std::basic_istream<charT, traits>&
 operator>>(std::basic_istream<charT, traits>& is, complex<ValueType>& z);
 
@@ -614,4 +615,3 @@ template <typename T> __host__ __device__ inline bool operator!=(const complex<T
 
 /*! \} // numerics
  */
-
--- a/thrust/complex.h
+++ b/thrust/complex.h
@@ -615,3 +615,4 @@ template <typename T> __host__ __device__ inline bool operator!=(const complex<T
 
 /*! \} // numerics
  */
+
--- a/thrust/detail/device_ptr.inl
+++ b/thrust/detail/device_ptr.inl
@@ -21,7 +21,6 @@
 
 #include <thrust/device_ptr.h>
 #include <thrust/device_reference.h>
-#include <iostream>
 
 #include <thrust/detail/type_traits.h>
 #include <thrust/iterator/iterator_traits.h>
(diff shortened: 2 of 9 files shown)


jaredhoberock:
Thanks Cris. I think adding this functionality is OK, but a couple questions:

1. Why an overload for `basic_ostream` and not just `ostream`?
2. Instead of `device_reference`, this function should probably be overloaded for `thrust::reference`. `device_reference` will pick it up as a descendant of `reference.

ccecka:
1. `ostream = basic_ostream<char, std::char_traits<char>>` and `wostream = basic_ostream<wchar_t, std::char_traits<wchar_t>>`. User-defined `char_traits` are rumored to exist...
2. I agree. The same should be done for `device_ptr` => `thrust::pointer`.
3. Hadn't considered it. At the moment, `op>>` doesn't work for any `thrust::device_reference<A>` because `operator A&()` can't be implemented. I suppose an impl would be of the form
```
template <typename C, typename T, typename X>
std::basic_istream<C,T>& operator>>(std::basic_istream<C,T>& s, thrust::device_reference<X>& x) {
  typename thrust::device_reference<X>::value_type t;
  s >> t;
  x = t;
  return s;
}
```
I guess it's nice for symmetry with `std::vector`.

I'll make these changes tomorrow.

jaredhoberock:
Thanks. If `operator>>` doesn't already work with `device_reference`, let's not try to make it right now.

ccecka:
Stream output operators for `thrust::pointer` and `thrust::reference` with Doxygen stubs for `thrust::device_ptr` and `thrust::device_reference`.

thrust/complex.h
```diff
@@ -614,4 +615,3 @@ template <typename T> __host__ __device__ inline bool operator!=(const complex<T
 
 /*! \} // numerics
  */
-
```
jaredhoberock: Please don't eliminate the newlines at the end of these files -- on some platforms, it causes a warning if the last line is not empty.

thrust/detail/device_ptr.inl
```diff
@@ -41,13 +40,6 @@ template<typename T>
   return ptr;
 } // end device_pointer_cast()
 
-// output to ostream
-template<class E, class T, class Y>
-  std::basic_ostream<E, T> &operator<<(std::basic_ostream<E, T> &os, const device_ptr<Y> &p)
```
jaredhoberock: I'd rather not introduce operator<< for pointer & device_ptr. Let's just fix the problems with reference for now.

thrust/detail/device_reference.inl
```diff
@@ -50,4 +50,3 @@ void swap(device_reference<T> &a, device_reference<T> &b)
 } // end swap()
 
 } // end thrust
-
```
jaredhoberock: no need to change this file

thrust/detail/pointer.h
```diff
@@ -178,7 +179,13 @@ template<typename Element, typename Tag, typename Reference, typename Derived>
     Element *get() const;
 }; // end pointer
 
+// Output stream operator
+template<typename Element, typename Tag, typename Reference, typename Derived,
+         typename charT, typename traits>
+std::basic_ostream<charT, traits> &
+operator<<(std::basic_ostream<charT, traits> &os,
```
jaredhoberock: Again, let's not provide this feature for pointer

jaredhoberock:
Thanks Cris, I added some notes inline on your commit.

thrust/detail/pointer.h
ccecka: It already exists for `thrust::device_ptr`, this simply promotes it to `thrust::pointer`.

jaredhoberock: OK, if it already exists, that's fine.

ccecka:
Clean diff.

jaredhoberock:
Thanks Cris, it looks good to me! I will test & merge your fixes tomorrow.

jaredhoberock:
The tests pass, so I can merge this change.

However, these fixes need to go into branch 1.8.2. Unfortunately, I can't change the target of this pull request -- can you target thrust:1.8.2 instead of thrust:master?

Issue #624, opened 2015-01-23
make unary_function/binary_function optional (C++11)


andrewcorrigan:
This is intended for the C++11 milestone:

Can requiring that `unary_function` and `binary_function` be the base class for a UnaryFunction or BinaryFunction in Thrust algorithms be made optional?

Not only does this remove the need for deriving from these types, I think (I'm just now starting to really learn C++11) that it also removes the need for writing constructors for small functors since they can instead be instantiated with aggregate initialization once the functors would no longer have a base class.

```
struct adds_a : public thrust::unary_function<int,int>
{
  int _a;
  adds_a(int a) : _a(a) {}
  int operator()(int i) const { return i + this->_a; }
};
...
adds_a(3);
```
becomes:
```
struct adds_a
{
  int _a;
  int operator()(int i) const { return i + this->_a; }
};
...
adds_a{3};
```

I've already made this change in my private copy of Thrust via `std::result_of` in `thrust/detail/type_traits/result_of.h`.
```
template<typename Signature>
  struct result_of
{
  typedef typename std::result_of<Signature>::type type;
};
```

andrewcorrigan:
Would you be interested in pull requests that can non-invasively start making use of C++11?

```
#if __cplusplus >= 201103L
template<typename Signature, typename Enable = void>
  struct result_of
{
  typedef typename std::result_of<Signature>::type type;
};
#else
template<typename Signature, typename Enable = void> struct result_of;

// specialization for unary invocations of things which have result_type
template<typename Functor, typename Arg1>
  struct result_of<
    Functor(Arg1),
    typename thrust::detail::enable_if<thrust::detail::has_result_type<Functor>::value>::type
  >
{
  typedef typename Functor::result_type type;
}; // end result_of
#endif
// specialization for binary invocations of things which have result_type
template<typename Functor, typename Arg1, typename Arg2>
  struct result_of<
    Functor(Arg1,Arg2),
    typename thrust::detail::enable_if<thrust::detail::has_result_type<Functor>::value>::type
  >
{
  typedef typename Functor::result_type type;
};
#endif

dachziegel:
@jaredhoberock How is your plan to continue the development Thrust with C+11 support?

dachziegel:
This would really help when using templated functors in combination with zip iterators, e.g:

    template <typename Tuple>
    struct Fun : public thrust::unary_function<Tuple, Tuple>
    {
	    __host__ __device__
	    Tuple operator()(Tuple p)
[...]

PR #671, opened 2015-07-08
remove the need to derive from thrust::unary_function/thrust::binary_function when c++11 is enabled

dachziegel:
see #624

### DIFF
--- a/thrust/functional.h
+++ b/thrust/functional.h
@@ -58,7 +58,9 @@ template<typename Operation> struct binary_traits;
  *  };
  *  \endcode
  *
- *  \note Inheriting from unary_function is optional if C+11 support is enabled.
+ *  \note Because C++11 language support makes the functionality of
+ *        \c unary_function obsolete, it's use is optional if C++11 language
+ *        features are enabled.
  *
  *  \see http://www.sgi.com/tech/stl/unary_function.html
  *  \see binary_function
@@ -96,7 +98,9 @@ struct unary_function
  *  };
  *  \endcode
  *
- *  \note Inheriting from binary_function is optional if C+11 support is enabled.
+ *  \note Because C++11 language support makes the functionality of
+ *        \c binary_function obsolete, it's use is optional if C++11 language
+ *        features are enabled.
  *
  *  \see http://www.sgi.com/tech/stl/binary_function.html
  *  \see unary_function
--- a/thrust/functional.h
+++ b/thrust/functional.h
@@ -58,9 +58,7 @@ template<typename Operation> struct binary_traits;
  *  };
  *  \endcode
  *
- *  \note unary_function is currently redundant with the C++ STL type
- *  \c std::unary_function. We reserve it here for potential additional
- *  functionality at a later date.
+ *  \note Inheriting from unary_function is optional if C+11 support is enabled.
  *
  *  \see http://www.sgi.com/tech/stl/unary_function.html
  *  \see binary_function
@@ -98,9 +96,7 @@ struct unary_function
  *  };
  *  \endcode
  *
- *  \note binary_function is currently redundant with the C++ STL type
- *  \c std::binary_function. We reserve it here for potential additional
- *  functionality at a later date.
+ *  \note Inheriting from binary_function is optional if C+11 support is enabled.
  *
  *  \see http://www.sgi.com/tech/stl/binary_function.html
  *  \see unary_function
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,33 @@
 #######################################
+#           Thrust v1.9.0             #
+#######################################
+
+Summary
+    TODO
+
+Breaking API Changes
+    None.
+
+New Features
+    TODO
+
+New Examples
+    TODO
+
+Other Enhancements
+    If C++11 support is enabled, functors do not have to inherit from thrust::unary_function/thrust::binary_function anymore when using them with thrust::transform_iterator.
+
+Bug Fixes
+    TODO
+
+Known Issues
+    TODO
+
+Acknowledgments
+    Thanks to Manuel Schiller for contributing a C++11 based enhancement regarding the deduction of functor return types.
+    
+
+#######################################
 #           Thrust v1.8.2             #
 #######################################
 
--- a/SConstruct
+++ b/SConstruct
@@ -321,9 +321,8 @@ def cc_compiler_flags(CXX, mode, platform, host_backend, device_backend, warn_al
   # workarounds
   result.extend(flags['workarounds'])
 
-  # select C++ standard, only set -std=... if we did not yet set it via nvcc
-  if device_backend != 'cuda':
-    result.extend(flags[cpp_standard])
+  # select C++ standard
+  result.extend(flags[cpp_standard])
 
   return result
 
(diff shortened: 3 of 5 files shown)


jaredhoberock:
You need to `#include <type_traits>` for `std::result_of`, right?

dachziegel:
* changed to `using`.
* included `<type_traits>`, it was working without when not using `nvcc`, e.g. `g++` and `clang` would compile without error.

Do we need a  unit test?

dachziegel:
regarding the unit tests: who can I enable `-std=c++11` when building the unit tests?

jaredhoberock:
I don't think the unit tests know about c++11. You would need to introduce a command line option to pass something like `-std=c++11` to the compiler.

There are dictionaries which map build options to compiler flags here:

https://github.com/thrust/thrust/blob/master/SConstruct#L28

I suggest introducing a new SCons `ListVariable` named `std` with the options `c++03` and `c++11`.

jaredhoberock:
I couldn't find any unit tests for `unary_function` or `binary_function`, so I think it's OK if there is no particular test for this feature. It would be a good idea to document that `thrust::unary_function` and `thrust::binary_function` are optional in C++11, though.

Regarding `using` -- now that we are not guarding its use with `__cplusplus`, I wonder if the implementation of `result_of` should be changed back to your original implementation. What do you think?

dachziegel:
regarding `using`: 

Do you think any compiler which defines `__cpp_lib_result_of_sfinae` would not support type aliases through `using`? I can hardly imagine this. But having a unit test then would make sense, if unit tests are run on all supported  compilers. What is your usual workflow regarding the execution of unit tests? Do you run on Linux, Mac OSX, Windows and different compilers?  Is there a list of supported compilers?

If we added a unit test for this, how would it be activated? Through that yet to be defined SCons `ListVariable` or using the same  `__cplusplus >= 201103L || defined(__cpp_lib_result_of_sfinae)` check done in code?

dachziegel:
Regarding documentation: Where should that hint go?

1) http://thrust.github.io/doc/classthrust_1_1permutation__iterator.html
2) http://thrust.github.io/doc/structthrust_1_1binary__function.html and http://thrust.github.io/doc/structthrust_1_1unary__function.html 

?

jaredhoberock:
It's difficult to predict the behavior of (buggy) compilers, so Thrust typically needs to be programmed as defensively as possible. Since it is a one-line change to change it back to a C++03-style `typedef`, we may as well do it.

Unit tests are run on a cross product of supported OSes and hardwares. Each particular OS version has a compiler version which is supported:

http://docs.nvidia.com/cuda/cuda-getting-started-guide-for-linux/index.html#system-requirements
http://docs.nvidia.com/cuda/cuda-getting-started-guide-for-mac-os-x/index.html#system-requirements
http://docs.nvidia.com/cuda/cuda-getting-started-guide-for-microsoft-windows/index.html#system-requirements

If we added a unit test for this feature, I would simply guard it with `__cplusplus >= 201103L`. However, because there are so many unit tests which are very expensive to build and execute, I do not think it is worth introducing a new one for what is a small quality of life feature.

For documentation, I think it would be sufficient to add a new note entry to `unary_function` and `binary_function` which indicates that they are optional with C++11. At the same time, I would eliminate the current note discussing redundancy with `std::unary_function`.

dachziegel:
Would you still like the possibility to switch enable c++11 for the existing unit tests through the scons variable?

jaredhoberock:
Finally, please add a new entry in the `CHANGELOG` for version 1.9.0. You can simply copy from 1.8.0 and fill in the sections with TODO. This feature should be mentioned under the "Other Enhancements" section. Also acknowledge yourself for a job well done!

Once we have some confidence that this change works, I'll merge it. Because the change is small, simply  building and running the example programs on your system should be sufficient.

jaredhoberock:
Yes, in order to test the example programs with this feature, it seems necessary to enable C++11 in the build.

In case it's not obvious, you can build & run the example programs automatically with

    $ scons run_examples

dachziegel:
You might add this command to the newly created wiki page.

jaredhoberock:
I've added a section for building & running the example programs.

dachziegel:
Could you please have a look at `SConstruct`? How can I avoid the warnings 

"cc1: warning: command line option '-std=c++11' is valid for C++/ObjC++ but not for C [enabled by default]" ?

dachziegel:
when not appending `-std=c++11` to `CCFLAGS` I don't get the warnings. Is the change to  `SConstruct` sufficient?

jaredhoberock:
No, I don't believe the change is sufficient. With your current changes, the C++11 build won't work unless `nvcc` is used as the compiler.

I think the warnings were probably due to `-Xcompiler -std=c++11` being passed to `nvcc`. Since `nvcc` supports `-std=c++11` directly, this option doesn't need to be wrapped with `-Xcompiler`.

Perhaps there's a way to remove `-std=c++11` from `CCFLAGS` when `nvcc` is used.

dachziegel:
I thought `nvcc` is mandatory when compiling the tests/examples? If this is assumption is wrong, then the information in the [Developer-Information wiki page](https://github.com/thrust/thrust/wiki/Developer-Information) I wrote is also incorrect:

> In order to run the tests, CUDA must be installed on your system because nvcc is used as the compiler driver.

dachziegel:
Unit tests passed on my Computer (Ubuntu 14.04):

    ================================================================                                                                                                                                                                  |
	KNOWN FAILURE: TestDeviceDeleteDestructorInvocation
	[testing/device_delete.cu:29]
	================================================================
	KNOWN FAILURE: TestScanWithLargeTypes
	[testing/scan.cu:504]
	================================================================
	Totals: 0 failures, 2 known failures, 0 errors, and 1374 passes.

Examples compiled and run after I fixed **cuda/async_reduce.cu**.

jaredhoberock:
I edited that part of the wiki page to indicate that that comment only applies to the CUDA backend. If you tests any other backend that does not use CUDA, the build will choose a normal C++ compiler.

dachziegel:
warnings do not occur anymore because `-std=c++11` is only set if not compiling via `nvcc`. Anything else to add/change?

thrust/functional.h
```diff
@@ -58,9 +58,7 @@ template<typename Operation> struct binary_traits;
  *  };
  *  \endcode
  *
- *  \note unary_function is currently redundant with the C++ STL type
- *  \c std::unary_function. We reserve it here for potential additional
- *  functionality at a later date.
+ *  \note Inheriting from unary_function is optional if C+11 support is enabled.
```
jaredhoberock: C+11 -> C++11

I think I would change the note to say something like:

"Because C++11 language support makes the functionality of \c unary_function obsolete, it's use is optional if C++11 language features are enabled."

Note that the `\c` makes Doxygen print the word in typewriter font.

thrust/functional.h
```diff
@@ -98,9 +96,7 @@ struct unary_function
  *  };
  *  \endcode
  *
- *  \note binary_function is currently redundant with the C++ STL type
- *  \c std::binary_function. We reserve it here for potential additional
- *  functionality at a later date.
+ *  \note Inheriting from binary_function is optional if C+11 support is enabled.
```
jaredhoberock: Same here

jaredhoberock:
Thanks Manuel, your changes look good to me. I found a couple of typos in `thrust/functional.h` and suggested a tweak to the note.

I will test your changes with MSVC and merge them if the tests pass on my system.

jaredhoberock:
Looks good, thanks for the hard work, Manuel!

Issue #677, opened 2015-07-14
implement thrust::unique using thrust::transform_iterator


dachziegel:
`thrust::unique` could be implemented using `thrust::transform_iterator`.
This avoids having to allocate temporary memory for the stencil and results in higher performance.

I implemented `thrust::unique_copy` using `thrust::transform_iterator` and benchmarked it against the original implementation:

see [thrust_unique_benchmark.cu](https://gist.github.com/dachziegel/191957d90aca8eafda32)

When running on Ubuntu 14.04, CUDA 7 and a GTX 680, this test case outputs:

    original: 3350407 microseconds
    new_impl: 2426929 microseconds
    new_impl is 27.5632 % faster

Would you be interested in a pull request?

jaredhoberock:
It sounds interesting. You might try avoiding [this single-element copy](https://gist.github.com/dachziegel/191957d90aca8eafda32#file-thrust_unique_benchmark-cu-L65) by sending along a counting iterator. It might make a difference to the performance.

dachziegel:
Yes, your idea (`new_impl_2`) is slightly faster and much simpler to implement: [thrust_unique_benchmark_v2.cu](https://gist.github.com/dachziegel/38764693902b71dcd861)

    original  : 3359968 microseconds
    new_impl  : 2424859 microseconds
    new_impl_2: 2410922 microseconds
    new_impl   is 27.8309 % faster than original
    new_impl_2 is 28.2457 % faster than original

I had something similar in mind, but did not like the `if` for the first element :)

Should I prepare the pull request?

jaredhoberock:
Fixed by #679

PR #679, opened 2015-07-14
implements #677

dachziegel:


thrust/system/detail/generic/unique.inl
```diff
@@ -101,13 +102,11 @@ __host__ __device__
   if(first == last)
```
jaredhoberock: i bet this check is not necessary with the new implementation

thrust/system/detail/generic/unique_by_key.inl
```diff
@@ -108,26 +109,23 @@ template<typename ExecutionPolicy,
                        BinaryPredicate binary_pred)
 {
   typedef typename thrust::iterator_traits<InputIterator1>::difference_type difference_type;
-  
+
   // empty sequence
   if(keys_first == keys_last)
     return thrust::make_pair(keys_output, values_output);
```
jaredhoberock: same here

jaredhoberock:
New implementation looks great. I'd see if those checks for empty sequences are no longer necessary because `copy_if` should be able to handle empty input.

Do these changes pass unit tests? I'll test them on Windows locally.

dachziegel:
Yes, unit tests are passed on Ubuntu 14.04 & CUDA 7.

dachziegel:
The latest commit  (28eed1828d520632f80ce708f399c7877c0393b4) also passed the unit tests.

jaredhoberock:
It passes on Windows. One last thing which I forgot before -- can you add an entry for this enhancement in the CHANGELOG as well as an acknowledgement? I'll merge at that point.

dachziegel:
done

jaredhoberock:
Thanks, great work!

Issue #699, opened 2015-09-08
Fix accidental s/omp/tbb/g in licenses text


daniel-j-h:
I was looking at some tbb code and stumbled upon this:

https://github.com/thrust/thrust/search?utf8=%E2%9C%93&q=ctbbliance

Seems like someone did a `s/omp/tbb/g` :P.

PR #700, opened 2015-09-08
License patch

zeryx:
corrected license text error as per ticket #699.


jaredhoberock:
You're absolutely right, nice catch and thanks for the fix!

zeryx:
No problem, anytime!

PR #705, opened 2015-11-17
Multiple patches

egaburov:


thrust/detail/allocator/temporary_allocator.inl
```diff
@@ -48,7 +48,8 @@ __host__ __device__
 #if !defined(__CUDA_ARCH__)
     throw thrust::system::detail::bad_alloc("temporary_buffer::allocate: get_temporary_buffer failed");
 #else
-    thrust::system::cuda::detail::terminate_with_message("temporary_buffer::allocate: get_temporary_buffer failed");
+    //thrust::system::cuda::detail::terminate_with_message("temporary_buffer::allocate: get_temporary_buffer failed");
+    thrust::system::cuda::detail::terminate();
 #endif
```
jaredhoberock: Why was this line changed to remove the message?

thrust/device_vector.h
```diff
@@ -97,7 +97,7 @@ template<typename T, typename Alloc = thrust::device_malloc_allocator<T> >
      *  \param v The \p device_vector to copy.
      */
     template<typename OtherT, typename OtherAlloc>
-    __device__
+    __device__ explicit
```
jaredhoberock: Why was explicit added here? It will likely break user code if device_vector cannot be assigned from other vectors.

jaredhoberock:
I got these warnings when building the example programs:

```
./thrust/system/cuda/detail/bulk/malloc.hpp(58): warning: address of a __shared__ variable "thrust::system::cuda::detail::bulk_::detail::s_data_segment_begin" cannot be directly taken in a host function
```

We should never commit changes that introduce new warnings. Can we try to avoid marking the functions in Bulk's malloc implementation with `__host__`? There might be another workaround to try. Logically, that malloc implementation cannot ever be used in a `__host__` function anyway.

PR #708, opened 2015-11-18
Split testing form a single uber-binary into a binary per test source file

egaburov:
Single binary can be built as:

`$ scons test.sort  [located in testing/sort]`
or
`$ scons test.cuda.sort [located in testing/backend/cuda/sort]`

The rest of functionality is preserved, `scons unit_tests` builds all unit tests, `$ scons run_unit_tests` or `$ scons run_tests` behaviour is preserved.


jaredhoberock:
I think I prefer the output of the old unit tester, where all results were collated into a single log. With this new organization, we receive multiple logs, one per executable. It makes it difficult to see at a glance whether there were failures.

Have you thought about a way that we could collate the individual tests' output for display?

egaburov:
We can pipe the output through another python script that collates the output, e.g.
`$ scons run_unit_tests | python collate_test_output.py`

Alternatively, I can what I did with DVS. Have a separate python script, in our case launched by `$ scons run_unit_tests`,  which globs binary path to collects tests binaries, runs them and collates output into a single report.

other suggestions are welcome.

jaredhoberock:
Yes, I was thinking of something along those lines.

I think this is sufficient for now. I'll merge this once you add the comments I mentioned.

egaburov:
done. next step, i will be working on a binary that collates output. alternatively, I will look into scons testing framework, you pointed out, that may do testing for us, similar to what `ctest` does in `cmake` package.

jaredhoberock:
OK. Instead of creating an additional program to collate the output, you might consider altering the unit tester framework such that the tests output a log that naturally collates when the logs are concatenated. It might be less complexity overall.

PR #709, opened 2015-11-19
Thrust clang

Artem-B:
As of r253389 clang can be used to compile CUDA source files.
Unlike nvcc, clang does not split original source into host/device. Instead it allows function overloading based on __host__/__device__ attribute. As a consequence, these attributes are no longer inheritable and function definitions must have attributes matching those of its declaration.

Clang also can not handle yet use of std::* on device side, so in order for examples to be compilable, I've replaced two instances with their global scope counterparts provided by CUDA headers.

With these patches all thrust examples are compilable with clang. Clang appears to miscompile histogram.cu, but output from other examples is identical to that from nvcc-compiled ones.


jaredhoberock:
Thanks for the contribution!

I noticed that you've disabled nvcc's `hd_warning_disable` & `nv_exec_check_disable`.

How does clang handle this case, which would cause nvcc to issue either warnings or errors:

```
template<class Function>
__host__ __device__
auto invoke(Function f) -> decltype(f())
{
  return f();
}

struct host_functor
{
  __host__
  void operator()(){}
};

struct device_functor
{
  __device__
  void operator()(){}
};

__global__
void kernel()
{
  return invoke(device_functor());
}

int main()
{
  invoke(host_functor());

  return 0;
}
```

Artem-B:
clang does not support these pragmas. That would not matter if not for the fact that
#pragma inside a macro results in syntax error in clang when compilation uses built-in preprocessor. 

That said, clang is less strict than nvcc, so your example produces no warnings or errors and appears to generate sensible code.

jaredhoberock:
Makes sense. Your changes look good to me, so I'll merge them.

Issue #710, opened 2015-11-19
Unit test build refactor causes can no longer handle host_backend=all device_backend=all


jaredhoberock:
The following `scons` command line causes an error on my system:

```
$ scons run_examples host_backend=all device_backend=all
scons: Reading SConscript files ...

scons: *** Two environments with different actions were specified for the same target: /home/jhoberock/dev/git/thrust-tot/testing/testframework.o
File "/home/jhoberock/dev/git/thrust-tot/testing/SConscript", line 72, in <module>
```

egaburov:
Argh horrible, let me look into this

egaburov:
Please check the fix. SCons behaves differently that I am used to, so please bear with me as I learn the system along the way. Thanks!

PR #713, opened 2015-11-20
Revert "Split testing form a single uber-binary into a binary per test source file"

jaredhoberock:
Reverts thrust/thrust#708


jaredhoberock:
Fixes #710

PR #721, opened 2015-12-24
add "thrust::zip_function" to apply ordinary functions on zipped iterators

schiller-manuel:
I often face the situation when I want to apply an ordinary function  on a zip_iterator, e.g:

```
struct sum3ints
{
    int operator()(int x, int y, int z)
    {
        return x+y+z;
    }
};

thrust::make_transform_iterator(thrust::make_zip_iterator(...), sum3ints());
```

In order to make it work, I would have to reimplement `sum3ints` to accept a single tuple instead of 3 arguments:

```
struct sum3ints_tuple
{
    template<typename Tuple>
    int operator()(const Tuple& t) { return thrust::get<0>(t) + thrust::get<1>(t) + thrust::get<2>(t); }
};
```

So I need to have 2 implementations of the same functionality.

I therefore created `thrust::zip_function` which is a wrapper around an ordinary function. This wrapper accepts a tuple and forwards the tuple elements as the parameters to the underlying function.

Since `thrust::zip_function` is fully generic, it also allows to apply `thrust::placeholders` on zip_iterators.

Some demo use cases are shown here: https://gist.github.com/dachziegel/de9b3610af4391aadc26
The output when running these demo use cases is shown here: https://gist.github.com/dachziegel/e11d3b641179412b0c3d


brycelelbach:
This is a cool idea, and I've love to have this. Some notes off the bat:

* I'm sure we have an implementation of something like index sequence somewhere. If not, it should go someplace more general, like type traits. 
* Right now, we still support C++03. It's fine for new features to be C++11 only, but if so, you need to put an #error for non-C++11 users when they include this.
* `std::forward` is not a device function. There's `nvstd::forward`; we should probably have our own though.
* Tests!

brycelelbach:

allisonvacanti:
Closing, as #1029 had the same purpose.

PR #722, opened 2015-12-24
add new "fancy" iterator: transform_output_iterator

dachziegel:
When answering [this stackoverflow question](http://stackoverflow.com/a/34424403/678093) I ported `transform_output_iterator` from [here](https://github.com/ldionne/boost-submissions/blob/master/boost/boost/iterator/transform_output_iterator.hpp) to thrust.

While `thrust::transform_iterator` acts as a "source" by applying a transformation on an `Iterator`, `thrust::transform_output_iterator` acts as "sink".
It allows  to apply a transform operation on a wrapped `OutputIterator` before actually writing to this `OutputIterator`. This can be used to post-process results from a previous thrust call.

### DIFF
--- a/thrust/iterator/detail/transform_output_iterator.inl
+++ b/thrust/iterator/detail/transform_output_iterator.inl
@@ -1,0 +1,53 @@
+#include <thrust/iterator/transform_output_iterator.h>
+#include <thrust/iterator/iterator_adaptor.h>
+
+namespace thrust
+{
+
+template <typename OutputIterator, typename UnaryFunction>
+  class transform_output_iterator;
+
+namespace detail 
+{
+
+template <typename UnaryFunction, typename OutputIterator>
+  class transform_output_iterator_proxy
+{
+
+  public:
+    __host__ __device__
+    transform_output_iterator_proxy(const OutputIterator& out, UnaryFunction fun) : fun(fun), out(out)
+    {
+    }
+
+    template <typename T>
+    __host__ __device__
+    transform_output_iterator_proxy operator=(const T& x)
+    {
+      *out = fun(x);
+      return *this;
+    }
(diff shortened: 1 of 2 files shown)


jaredhoberock:
Thanks. This is a good addition because it enables new usage that is not easily realized without it. I'm particularly interested in the opportunity for implementing an inner join type of algorithm through a composition of `set_intersection` and `transform_output_iterator`.

Because this is a new feature, it requires documentation, unit tests, and ideally incorporation into an example program.

@egaburov, could you handle the additional tasks of documentation & testing? I'll merge this pull request once they are finished.

egaburov:
@dachziegel  It appears this transform iterator doesn't helps to implement an inner join type of algorithm. The correct result in the SO example [1] is due entirely to coincidence in the input data: the offset of the same key in both tables is the same.

For example, if the second key-value pair in the Left table is removed, the result should remain unchanged, because the key "2" is not found in the Right table. Yet, the example code produces different result:

    $ nvcc x.cu -std=c++11 -run
    result_keys:    1       4       5       6
    result_values:  6       1       2       4


It appears `thrust::set_intersection_by_key` accesses `zipped_values` via an offset computed only from the right table, which can explain the difference in the result. It is not clear now how transform_output_iterator can help to implement an inner join. 

Could we think of other good use of such output iterator?

.. [1] http://stackoverflow.com/questions/34370201/combining-two-lists-by-key-using-thrust/34424403#34424403

jaredhoberock:
@dachziegel To add to what Evghenii is saying -- does the solution posted in the Stack Overflow thread solve the original problem comprehensively?

If it does, what are we missing? If does not, is there a simple fix we could apply which yields a comprehensive solution?

PR #724, opened 2015-12-28
provide NVVMIR_LIBRARY_DIR environment variable to nvcc

dachziegel:
If a system does not use a `nvcc.profile` file but sets `NVVMIR_LIBRARY_DIR` as an environment variable instead, building thrust tests fails with:

```
 nvcc fatal : Path to libdevice library not specified
```

This patch provides `NVVMIR_LIBRARY_DIR` to `nvcc` if this environment variable is set.


brycelelbach:
Sure, sounds good. We are moving to CMake eventually.

PR #729, opened 2016-01-13
Added rough support for compiling thrust with clang.

Artem-B:
* added compiler name to object directory name
* added SCons option: cuda_compiler=(nvcc|clang)
* path to clang must be specified via CLANG_PATH environment variable.

NOTE: Compiling CUDA with clang requires a recent build from development branch of clang-3.8 for Linux available here: http://llvm.org/apt/.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -106,23 +106,15 @@ linker_to_flags = {
   'clang++'  : clang_linker_flags
 }
 
-
-def cuda_installation():
+def cuda_installation(env):
   """Returns the details of CUDA's installation
   returns (bin_path,lib_path,inc_path,library_name)
   """
 
-  # determine defaults
-  if os.name == 'nt':
-    bin_path = 'C:/CUDA/bin'
-    lib_path = 'C:/CUDA/lib'
-    inc_path = 'C:/CUDA/include'
-  elif os.name == 'posix':
-    bin_path = '/usr/local/cuda/bin'
-    lib_path = '/usr/local/cuda/lib'
-    inc_path = '/usr/local/cuda/include'
-  else:
-    raise ValueError, 'Error: unknown OS.  Where is nvcc installed?'
+  cuda_path = env['cuda_path']
+  bin_path = cuda_path + '/bin'
+  lib_path = cuda_path + '/lib'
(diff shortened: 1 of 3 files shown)


SConstruct
```diff
@@ -355,6 +347,12 @@ def nv_compiler_flags(mode, device_backend, arch, cdp):
   
   return result
 
+def clang_compiler_flags(mode, arch):
+  """Returns a list of command line flags specific to nvcc"""
```
jaredhoberock: "nvcc" -> "clang"

SConstruct
```diff
@@ -106,23 +106,15 @@ linker_to_flags = {
   'clang++'  : clang_linker_flags
 }
 
-
-def cuda_installation():
+def cuda_installation(env):
```
jaredhoberock: For consistency, instead of receiving an `env` here, could we take the name of the CUDA compiler, in the same way that `omp_installation` receives the name of the `CXX` compiler?

jaredhoberock:
Thanks Artem, this looks helpful. I made some suggestions for making your change a little more consistent with the way the rest of the build works. The build is getting complex, so I'd like to keep things as consistent as possible. What do you think?

SConstruct
Artem-B: omp_installation is not a good example, IMO. All it's used for is to figure out library name which it does derive from compiler name.
In this case I can either pass cuda_path instead of env or pass nothing, return paths relative to CUDA installation and then stitch them into absolute path where it's used. I don't think either of these options improves things.

SConstruct
Artem-B: Done.

Artem-B:
On a side note, did you consider switching to something a bit less esotheric than SCons? CMake, perhaps?

SConstruct
jaredhoberock: Thanks for pointing that out. I see now that the other like functions such as `tbb_installation` take an `env`, so `omp_installation` is really the odd one out. So I think your change to `cuda_installation` is OK.

dachziegel:
@jaredhoberock I'd also like a CMake build :+1:

PR #733, opened 2016-01-27
Add fake #includes to fix SCons dependencies in thrust/system/detail/…

jlebar:
…adl headers.

SCons figures out dependencies by textually scanning header files.  This
means that it misses dependencies of the form e.g.

    #define foo bar.h
    #include foo

Unfortunately this is exactly what we do in the adl headers.

In the case of the adl headers, we're using #defines to switch between a
few possible header files, so it's fine if SCons simply generates a
dependency on all of the files that we might depend on.  To accomplish
this, we add #includes for all files that we might include, but stick
those #includes inside an #if 0.  This way the includes are visible only
to SCons.

### DIFF
--- a/thrust/system/detail/adl/adjacent_difference.h
+++ b/thrust/system/detail/adl/adjacent_difference.h
@@ -24,6 +24,16 @@
 
 #include <thrust/system/detail/sequential/adjacent_difference.h>
 
+// SCons can't see through the #defines below to figure out what this header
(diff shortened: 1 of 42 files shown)


jaredhoberock:
Thanks for the patch -- this limitation has always been annoying. I think it's OK to do this kind of fakery because we do something similar elsewhere to accommodate Doxygen.

PR #734, opened 2016-01-30
Make type traits work with clang.

jlebar:
See http://libcxx.llvm.org/type_traits_design.html for the list of intrinsics supported by clang.


jaredhoberock:
Looks good. Thanks, Justin!

PR #735, opened 2016-02-02
fix window warnings

egaburov:


jaredhoberock:
If you look at the diff, it looks like your change touched every line in these files. Probably what happened is that Windows introduced new line endings or whitespace.

Could you resubmit your change such that the diff is limited to the edits of interest?

egaburov:
Looks correct now.

jaredhoberock:
Thanks Evgehnii, looks good to me. Since these changes are in the Bulk library, we need to keep them in sync. Could you submit a corresponding pull request here:

https://github.com/jaredhoberock/bulk

I'll merge both at the same time.

egaburov:
Done.

PR #736, opened 2016-02-02
Pull isinf/isnan/signbit and isfinite from std namespace

Artem-B:
... if we're compiling with clang which (as of r258880) provides device-side wrappers for them.

This is needed to provide clang with access to both host and device-side variants of those functions.
Global scope will not have host-side variants available if they were implemented as preprocessor macros.

### DIFF
--- a/thrust/detail/complex/c99math.h
+++ b/thrust/detail/complex/c99math.h
@@ -100,7 +100,7 @@ __host__ __device__ inline int isfinite(double x){
 
 #else
 
-#  ifdef __CUDACC__
+#  if defined(__CUDACC__) && !(defined(__CUDA__) && defined(__clang__))
 
 // sometimes the CUDA toolkit provides these these names as macros,
 // sometimes functions in the global scope


egaburov:
we have to be careful with `defined(__clang__)` because on OS X nvcc uses clang as a host compiler, and thus defined(__clang__) returns always true. Before committing, I suggest to verify that this doesn't introduces errors and new warnings.

Artem-B:
Hence (defined(__CUDA__) && defined(__clang__))
__CUDA__ is set by clang (very recent one at that) only when we compile CUDA files. It will not be set if you use clang as a host compiler used by nvcc because in that case it will be compiling C++ inputs.

jaredhoberock:
Thanks Artem & Evghenii.

Evghenii, could you test this change on a OS X install of nvcc? If you are satisfied with Artem's change, I will merge it.

egaburov:
Should be safe to merge.

PR #737, opened 2016-02-02
Fix osx compilation

egaburov:
Added defined(__CUDA__) guard to enable `#pragma` when host compiler is clang.

### DIFF
--- a/testing/unittest/assertions.h
+++ b/testing/unittest/assertions.h
@@ -29,8 +29,13 @@ namespace unittest
 
 static size_t MAX_OUTPUT_LINES = 10;
 
+#ifndef __APPLE_CC__
 static double DEFAULT_RELATIVE_TOL = 1e-4;
 static double DEFAULT_ABSOLUTE_TOL = 1e-4;
+#else
+#define DEFAULT_RELATIVE_TOL (1e-4)
+#define DEFAULT_ABSOLUTE_TOL (1e-4)
+#endif
 
 template<typename T>
   struct value_type
--- a/testing/unittest/assertions.h
+++ b/testing/unittest/assertions.h
@@ -29,13 +29,8 @@ namespace unittest
 
 static size_t MAX_OUTPUT_LINES = 10;
 
-#ifndef __APPLE_CC__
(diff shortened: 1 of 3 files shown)


egaburov:
OS X clang still generates tons of warnings if compiled with Wall=yes. It is quiet though when built with Wall=no.

jaredhoberock:
Are the warnings related to this pull request?

jaredhoberock:
The change looks good to me.

Also, a couple of things:

  1. Since this change also touches Bulk, we'll need a corresponding pull request
  2. Do you want to target this at 1.8.3 or just master? My vote is for 1.8.3, since it's always nice to eliminate warnings in point releases, but it's up to you.

egaburov:
> Are the warnings related to this pull request?
No, the remaining warnings are not related to this PR. W/o #pragma nv_exec_check_disable, there is a error

    error: calling a __device__ function("singleton_on_chip_allocator") from a __host__ __device__ function("construct") is not allowed

which breaks compilation. This PR is requirde to fix compilation on Mac.

I will submit PR to Bulk.


1.8.3 has only defined(__CUDACC__), thus it will not break compilation with nvcc on Mac, unlike master; fwiw, 1.8.3 still generates tons of warnings with Wall=yes, but these are not related to this PR.

jaredhoberock:
OK, so it sounds like 1.8.3 does not require this change. We can pursue the other warnings you mention elsewhere.

egaburov:
PR to Bulk is submitted. I think it makes little sence to backport changes to 1.8.3 since 1.8.3 can't built with clang, afaik

jaredhoberock:
Hrm, I think the Bulk PR I just merged had a mistake.

Shouldn't the clang check go here:

https://github.com/jaredhoberock/bulk/blob/master/bulk/detail/config.hpp#L27

instead of here:

https://github.com/jaredhoberock/bulk/blob/master/bulk/detail/config.hpp#L44

At least, that's where your check is in this PR.

egaburov:
Bugger. My mistake. Lemme resubmit PR against current master.

PR #740, opened 2016-02-02
Added managed memory allocator example

sdalton1:


jaredhoberock:
Thanks Steve. I think you can eliminate much of the members of the allocator. I think all that is necessary is `.allocate()` and `.deallocate()`. Take a look at `custom_temporary_allocation.cu` to see what I mean.

sdalton1:
Thanks Jared. I will close this request for now. I need to experiment with managed memory a bit more and come up with a motivating example to use it.

egaburov:
This managed memory allocator is very useful, when used in combination with device_vector. The device_vector iterator then can be passed to both host and device algorithms.

PR #743, opened 2016-02-04
Fix hd warning on destructor

egaburov:
Fixes https://github.com/thrust/thrust/issues/741


thrust/device_vector.h
```diff
@@ -69,6 +69,9 @@ template<typename T, typename Alloc = thrust::device_malloc_allocator<T> >
     device_vector(void)
       :Parent() {}
 
+    __host__
```
jaredhoberock: This new destructor needs a Doxygen comment. You can use the Doxygen for the default constructor for an example.

thrust/host_vector.h
```diff
@@ -69,6 +69,9 @@ template<typename T, typename Alloc = std::allocator<T> >
     host_vector(void)
       :Parent() {}
 
+    __host__
```
jaredhoberock: This constructor needs Doxygen too.

jaredhoberock:
Thanks Evghenii. I'll merge once the comments are introduced.

thrust/device_vector.h
```diff
@@ -69,6 +69,12 @@ template<typename T, typename Alloc = thrust::device_malloc_allocator<T> >
     device_vector(void)
       :Parent() {}
 
+    /*! Define an empty destructor to explicitly specify
```
jaredhoberock: It's fine to leave a note to say why we're defining these destructors (i.e. to work around an nvcc warning), but I'm not sure it makes sense for it to show up in the Doxygen-generated documentation. (The `/*!` syntax causes the comment to be visible via Doxygen.)

If a user read this in the documents on the web, the stuff about execution space would probably just cause confusion. So, I think the user-visible documentation should just say something similar to what is said here: http://en.cppreference.com/w/cpp/container/vector/~vector.

egaburov:
Is this better? I copy-pasted documentation from vector_base.h

jaredhoberock:
Good idea. That looks great.

jaredhoberock:
Ah, one last thing. Since we're fixing a warning, let's merge this into 1.8.3 to kill it sooner. Could you retarget the PR?

egaburov:
PR is retargeted to 1.8.3 in https://github.com/thrust/thrust/pull/744. Closing

PR #746, opened 2016-02-19
range_view example

egaburov:
Construct a range_view from a pair of iterators with with a vector-like interface.

### DIFF
--- a/examples/cuda/range_view.cu
+++ b/examples/cuda/range_view.cu
@@ -4,6 +4,9 @@
 #include <thrust/execution_policy.h>
 #include <iostream>
 
+// This example demonstrates implementaiton of range_view containers
+// which stores an iterator range with Vector-like interface
+
 template<class Iterator>
 class range_view
 {
@@ -112,6 +115,8 @@ public:
 
 };
 
+// This helper function reates a range_view from iterator and the number of
+// elements
 template <class Iterator, class Size>
 range_view<Iterator>
 __host__ __device__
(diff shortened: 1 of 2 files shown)


examples/cuda/range_view.cu
```diff
@@ -0,0 +1,239 @@
+#include <thrust/device_vector.h>
+#include <thrust/for_each.h>
+#include <thrust/iterator/counting_iterator.h>
+#include <thrust/execution_policy.h>
+#include <iostream>
+
+// This example demonstrates implementaiton of range_view containers
```
jaredhoberock: "implementaiton" -> "implementation"

jaredhoberock:
It looks fine to me. I think I would flesh out the prose at the top explaining what the example is meant to demonstrate and why one would want to write their code this way. Since some programmers won't be familiar with the idea of ranges or views, a small introduction might be appreciated.

egaburov:
is it okay now?

jaredhoberock:
Thanks, it looks a lot better. What do you think of this edit?

> This example demonstrates the use of a view: a non-owning wrapper for an iterator range which presents a container-like interface to the user.
> 
> For example, a view of a device_vector's data can be helpful when we wish to access that data from a __device__ function. Even though device_vectors are not accessible from __device__ functions, the range_view class allows us to access and manipulate its data as if we were manipulating a real container.

jaredhoberock:
Oh, one last thing: please note this new example in the 1.8.3 CHANGELOG.

egaburov:
Thanks Jared, your version reads better. I've replaced the prose with it and added 1.8.3 Changelog

PR #748, opened 2016-02-23
Make on_chip_cast a nop under clang.

jlebar:
This function relies on UB, which causes clang to miscompile it.  It's
not clear how to get equivalent functionality without UB, so since this
is just an optimization, make it a no-op.

With this change, clang (with some changes still under review) runs all
the thrust tests with no failures!

### DIFF
--- a/thrust/system/cuda/detail/bulk/malloc.hpp
+++ b/thrust/system/cuda/detail/bulk/malloc.hpp
@@ -38,9 +38,24 @@ inline __device__ bool is_on_chip(void *ptr)
 template<typename T>
 inline __device__ T *on_chip_cast(T *ptr)
 {
+#if defined(__CUDA__) && defined(__NVCC__) && !defined(__clang__)
+  // The below is UB in three ways:
+  //  * s_begin is not defined anywhere, so using it is an ODR violation.
+  //  * Pointer arithmetic is not defined to wrap, so (ptr - s_begin) + s_begin
+  //    is not necessarily ptr.
+  //  * Given a base pointer p, it's illegal to compute an address that's beyond
+  //    1 + the allocated size of p.  So in particular, if p is unallocated (as
+  //    here), it's illegal to do *any* pointer arithmetic on p.
+  //
+  // Some of this UB causes clang to miscompile this function.  Since it's just
+  // an optimization, enable it only for nvcc for now.  We can revisit this if
+  // the performance impact is large.
   extern __shared__ char s_begin[];
   void *result = (reinterpret_cast<char*>(ptr) - s_begin) + s_begin;
   return reinterpret_cast<T*>(result);
+#else
+  return ptr;
+#endif
 } // end on_chip_cast()
 
 


thrust/system/cuda/detail/bulk/malloc.hpp
```diff
@@ -38,9 +38,24 @@ inline __device__ bool is_on_chip(void *ptr)
 template<typename T>
 inline __device__ T *on_chip_cast(T *ptr)
 {
+#ifndef __clang__
```
egaburov: Please make it #if !(defined (__clang__) && defined(__CUDA__)). If clang is  a host compiler, the wrong code path will be taken with  nvcc.

jaredhoberock: I think I would make the UB code depend on NVCC rather than !clang. The desired effect seems to depend on an optimization that NVCC applies.

jaredhoberock:
Very nice work Justin, and thanks for the detailed analysis on UB. Congrats on passing the test suite! I think the only change I would suggest would be to invert the test the `#if` applies.

jlebar:
Updated as described.  I like this even better!

jaredhoberock:
Thanks Justin!

thrust/system/cuda/detail/bulk/malloc.hpp
```diff
@@ -38,9 +38,24 @@ inline __device__ bool is_on_chip(void *ptr)
 template<typename T>
 inline __device__ T *on_chip_cast(T *ptr)
 {
+#if defined(__CUDA__) && defined(__NVCC__) && !defined(__clang__)
```
egaburov: nvcc doesn't define `__CUDA__` afaik, so this path won't be taken and `nvcc` will fall back to `#else`. We need something like that

    #ifdef __NVCC__  

 or

    #if !(defined(__CUDA__) && defined(__clang__))

but not 

    #if defined(__CUDA__) && defined(__NVCC__) && !defined(__clang__)

If clang is a host compiler both `__NVCC__` and `__clang__` will be defined.

jlebar:
(Alternatively we can just make THRUST_DEVICE_COMPILER not lie and say that clang is nvcc.  I guess that's a maintenance question for you all, though, because it means more complicated ifdefs in a bunch of places.)

thrust/system/cuda/detail/bulk/malloc.hpp
jlebar: Yikes, sorry about that.  I checked the capitalization of `__NVCC__` but it didn't even occur to me that nvcc wouldn't define `__CUDA__`.

> We need something like

    #ifdef __NVCC__  

except this doesn't work because clang defines `__NVCC__` (which I'm not at all happy about, but that's another story)

> or

    #if !(defined(__CUDA__) && defined(__clang__))

except we said we wanted this to trigger only if nvcc, and this triggers if not clang.

So I think we may *actually* want to detect "really NVCC" with something like

    #if defined(__NVCC__) && !(defined(__CUDA__) && defined(__clang__))

I'll send a patch if we agree this is right?

egaburov: This looks good to me and I'm okay with this ifdef for now. I am not sure what you suggest at the bottom though, but would it make sense for clang to define __clang_cuda__ macro, or alike; or is this something better be done in Thrust?

egaburov: p.s.: nvcc defines `__CUDACC__`  when compiling CUDA files [1]

[1] http://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/#nvcc-identification-macro

jlebar:
>  I am not sure what you suggest at the bottom though, but would it make sense for clang to define 
> __clang_cuda__ macro, or alike; or is this something better be done in Thrust?

In compiler.h, we have

    #if defined(__CUDACC__)
    [...]
    #if THRUST_HOST_COMPILER == THRUST_HOST_COMPILER_CLANG
    // CUDA-capable clang should behave similar to NVCC.
    #if defined(__CUDA__)
    #define THRUST_DEVICE_COMPILER THRUST_DEVICE_COMPILER_NVCC
    #else
    #define THRUST_DEVICE_COMPILER THRUST_DEVICE_COMPILER_CLANG
    #endif

which, as I read it, means that we can't use THRUST_DEVICE_COMPILER to tell the difference between nvcc and clang.

> This looks good to me and I'm okay with this ifdef for now. 

I'll spin a patch.

jaredhoberock:
I agree that compiler.h should be honest about `THRUST_DEVICE_COMPILER` when a CUDA-capable clang is present. We can work that out later, though, and just patch the problem discussed in this PR in the meantime.

PR #749, opened 2016-02-24
Tweak condition in on_chip_cast used for detecting specifically nvcc …

jlebar:
…(not clang).

This was added in b59890f, but the condition was wrong -- nvcc doesn't
declare __CUDA__.


jaredhoberock:
Thanks Justin! @egaburov, I'll merge if you're satisfied with the change.

egaburov:
LGTM. Thanks

PR #752, opened 2016-03-01
Transform output iterator from #722

egaburov:
PR with adapted https://github.com/thrust/thrust/pull/722 to support `transform_output_iterator` in Thrust with algorithms.

Includes: 

-    transform_output_iterator from https://github.com/thrust/thrust/pull/722
-    example
-    test
-    licence
-    documentation & comments


jaredhoberock:
Thanks Evghenii, the code looks good. I think the example program needs a bit of introductory prose to explain what's going on. You can draw from another example that is primarily meant to demonstrate a fancy iterator.

I wonder if there's a better use case for `transform_output_iterator` than using `thrust::gather`. It seems like you would get the same effect from this example if you used a `transform_iterator` on the input. That might be something to note in a comment in the example.

Can we think of some use of `transform_output_iterator` which is not possible with `transform_iterator`?

Also, we should update the CHANGELOG and credit @dachziegel for this addtion.

egaburov:
Done. I modified the example to use transform_output_iterator to apply a function on multiple gathered elements before storing result.

PR #753, opened 2016-03-07
improve naming to represent intent; use `std::generate`

xtofl:


examples/bounding_box.cu
```diff
@@ -66,21 +66,17 @@ int main(void)
...
   // generate some random points in the unit square
-  for(size_t i = 0; i < N; i++)
-  {
-      float x = u01(rng);
-      float y = u01(rng);
-      points[i] = point2d(x,y);
-  }
+  std::generate(points.begin(), points.end(), [&] {
+    auto x = u01(rng);
+    auto y = u01(rng);
```
brycelelbach: Let's avoid auto here, we're not requiring C++11 yet.

xtofl: If we're not requiring C++11, we can't use this lambda either, can we?

brycelelbach:

xtofl:
That was easy.  Rebased/pushed.

xtofl:
Is something happening to this pull request?  Should I close it?

brycelelbach:
Hey, we will take a look at it, but we are a little backed up with our release cycle right now

Sent from Nine<http://www.9folders.com/>
________________________________

xtofl:
No problem.  It is, after all, only a tiny first step in pouring 'modern C++' over the thrust examples.

I think it's part of a bigger effort to take on the whole example code.  To me, it's important that examples breath good style, so that when I point people at them, they fullful a larger educational purpose than just showcasing the library.  I'll make an Issue (#1041) out of that, as to formalize it.

PR #756, opened 2016-03-15
fix: win32 copy when cuda execution policy is specified

egaburov:


thrust/system/cuda/detail/trivial_copy.inl
```diff
@@ -66,6 +66,16 @@ template<typename System1,
   return cudaMemcpyHostToDevice;
 } // end cuda_memcpy_kind()
 
+#if defined(_WIN32) && !defined(_WIN64)
```
jaredhoberock: I think this overload would make sense no matter the OS or bitness, right? In order to keep the conditional compilation to a minimum, could we safely remove this `#if`?

egaburov: Not quite: CUDA supports 32-bit mode only on Windows, ergo windows-specific fix.

We can't remove #if because of the  possible use  thrust::copy(thrust::cuda::par, raw_host_ptr, raw_host_ptr+n, raw_device_ptr). The cudaMemcpyDefault in 64-bit will figure out correct src/destnation memory space and will do host->device, copy.

jaredhoberock: Hmm. I misunderstood what the `#if` was guarding. I thought it was surrounding the entire function. Now I see that it is surrounding the implementation.

I'm afraid I don't understand the implementation of this overload. If we're copying from CUDA -> CUDA, why wouldn't the result of this function always be `cudaMemcpyDeviceToDevice`?

jaredhoberock:
I think I understand now. You've updated the implementation since your first commit and moved the `#if` to only apply to the definition of `cuda_memcpy_kind()`. You also unconditionally call it below. That's good, because I was going to suggest that change.

The last enhancement I would recommend is a comment inside of `cuda_memcpy_kind` that explains that we're assuming that any raw pointers encountered by `cuda_memcpy_kind` on Win32 are going to be device pointers and that host <-> device copies with `cuda::par` basically aren't supported on this platform.

egaburov:
Okay, added comment to explain conditional compilation. This should make it clear why it is done this way.

It is ironic that w/o this fix, the very first example on https://thrust.github.io fails to run on Win32.

PR #761, opened 2016-03-22
Add additional __host__ __device__ attributes.

jlebar:
clang is stricter about requiring defs' and decls' attributes to match
than nvcc.  Currently clang doesn't care if you have a `__host__`
`__device__` decl and an unattributed def, but that is likely to change
soon, as a side-effect of supporting --relaxed-constexpr.

This patch also cleans up some whitespace.


jaredhoberock:
Thanks Justin, looks good to me.

PR #763, opened 2016-03-23
Simplify ifdef for detecting nvcc in malloc.hpp.

jlebar:
Clang tip of trunk (which is all we support at the moment) now no longer
defines `__NVCC__`, so this ifdef can be simplified.


jaredhoberock:
Thanks Justin.

@egaburov Does this look OK to you?

egaburov:
LGTM

PR #765, opened 2016-03-30
remove redundant hd_warning_disable

egaburov:


thrust/detail/config/exec_check_disable.h
```diff
@@ -25,7 +25,6 @@
 #if defined(__CUDACC__)
 
 #define __thrust_exec_check_disable__ \
```
jaredhoberock: We should also eliminate this trailing slash

thrust/system/cuda/detail/bulk/detail/config.hpp
```diff
@@ -27,7 +27,7 @@
 #if defined(__CUDACC__)
 #  ifndef __bulk_hd_warning_disable__
 #    define __bulk_hd_warning_disable__ \
```
jaredhoberock: This one too.

thrust/detail/config/exec_check_disable.h
jaredhoberock: Sorry, I'm not reading the diff correctly. Looks good.

PR #767, opened 2016-04-01
added guards based on cuda version

egaburov:
based on comment https://github.com/thrust/thrust/issues/688#issuecomment-204121313


jaredhoberock:
Thanks, it looks good to me. To simplify future merges with `master`, I will go ahead and also pull this into `master`. If we want to eliminate the older `hd_warning_disable` in the future, we can do so in the `master` branch.

PR #770, opened 2016-04-03
Fix UB in float3_optimization performance test.

jlebar:
The rotate_tuple kernel was returning references to stack memory.  nvcc
didn't notice or care, but clang did, and optimized away the whole
function.  I suppose it was equally correct, and it was indeed faster.
:)

Fixes thrust/thrust#769.


jaredhoberock:
Looks good to me, thanks Justin!

PR #772, opened 2016-04-07
Taught thrust::complex to provide an empty constructor.

stephen-sorley:
This patch allows thrust::complex objects to be created with **shared**, **device** or **global** scope.

This change fixes a compiler error we encountered (dynamic initialization not supported) when trying to use thrust::complex in an OptiX ray payload struct.

### DIFF
--- a/thrust/complex.h
+++ b/thrust/complex.h
@@ -67,13 +67,19 @@ public:
 
   /* --- Constructors --- */
 
+  /*! Construct an uninitialized complex number.
+   *
+   */
+  inline __host__ __device__
+  complex();
+
   /*! Construct a complex number from its real and imaginary parts.
    *
    *  \param re The real part of the number.
    *  \param im The imaginary part of the number.
    */
   inline __host__ __device__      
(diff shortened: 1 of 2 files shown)


jaredhoberock:
Thanks for the suggestion. I don't think we can apply this change because it would alter the behavior of default-constructed `complex` objects.

While working around these kinds of problems in the past, I've found it useful to introduce an `unintialized<T>` wrapper for the type of interest. The idea is that `uninitialized<T>` is a container for an object whose constructor and destructor are ignored. The code in [this repository](https://github.com/jaredhoberock/uninitialized) may be helpful to you.

stephen-sorley:
Thanks for the quick response!

I would argue that complex scalar types in a numerical library should have the same semantics as the corresponding real types - ideally you'd want the ability to substitute in complex<float> anywhere that float is currently being used.  However, given that std::complex in C++ is defined in the same way (automatically zero-initialized), I can certainly understand why you'd want to keep the existing behavior.

Thanks for the tip on uninitialized<T>, I'll give it a try.

PR #777, opened 2016-04-14
Dereference iterator via get_value(derived_cast(exec),iterator)

egaburov:
Fixes https://github.com/thrust/thrust/issues/776

This allows peeking into the device data when exec is device policy
and the iterator is a raw device pointer, thus fixing
{min,max,minmax}_element in such setup.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -6,12 +6,11 @@ Summary
     Small bug fixes
 
 New Examples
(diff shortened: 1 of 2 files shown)


jaredhoberock:
Thanks, it looks good. Could you also please include a note in the CHANGELOG about this fix (as well as your previous fix which I think we forgot to mention).

PR #783, opened 2016-04-29
introduce get_iterator_value

egaburov:
This fixes regression https://github.com/thrust/thrust/issues/780 introduced in https://github.com/thrust/thrust/pull/777 without regressing the origin fix https://github.com/thrust/thrust/issues/776

@sdalton1 can you please verify this fixes your issues?

@jaredhoberock While the fix being reviewed, I will work on adding unit test coverage for this functionality.  Don't merge just yet.

### DIFF
--- a/thrust/detail/get_iterator_value.h
+++ b/thrust/detail/get_iterator_value.h
@@ -1,0 +1,51 @@
+#pragma once
+/*
+ *  Copyright 2008-2016 NVIDIA Corporation
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+#pragma once
+
+#include <thrust/detail/config.h>
+
+namespace thrust {
+namespace detail {
+
+// get_iterator_value specialization on iterators
+// --------------------------------------------------
+// it is okay to dereference iterator in a usual way
+template<typename DerivedPolicy, typename Iterator>
+__host__ __device__
+typename thrust::iterator_traits<Iterator>::value_type
+get_iterator_value(thrust::execution_policy<DerivedPolicy> &, Iterator it)
+{
+  return *it;
+} // get_iterator_value(exec,Iterator);
+
+// get_iterator_value specialization on pointer
+// ----------------------------------------------
+// we can't just dereferecne a pointer in usual way, because
+// it may point to a location in the device memory. 
+// we use get_value(exec,pointer*) function
+// to perform a dereferencing consistent with the execution policy
+template<typename DerivedPolicy, typename Pointer>
+__host__ __device__
+typename thrust::detail::pointer_traits<Pointer*>::element_type 
+get_iterator_value(thrust::execution_policy<DerivedPolicy> &exec, Pointer* ptr)
+{
+  return get_value(derived_cast(exec),ptr);
+} // get_iterator_value(exec,Pointer*)
+
+} // namespace detail
+} // namespace thrust
--- a/thrust/detail/get_iterator_value.h
+++ b/thrust/detail/get_iterator_value.h
@@ -15,8 +15,6 @@
  *  limitations under the License.
  */
 
-#pragma once
-
 #include <thrust/detail/config.h>
 
 namespace thrust {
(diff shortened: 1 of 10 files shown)


jaredhoberock:
Thanks Evghenii. Since this new function isn't a customization point (we don't want users to try to customize its behavior, and we don't want to ourselves, either), let's not organize it into `thrust/system/detail/generic/`.

How about `thrust/detail/get_iterator_value.h`?

Also, the way we should invoke it is:

    detail::get_iterator_value(...)

Instead of

    get_iterator_value(...)

Unlike `get_value()`, we don't want to call it through ADL.

egaburov:
Thanks for the explanation. I now understand better the directory hierarchy. I've moved the implementation to thrust/detail/get_iterator_value.h , and add unit testing for issues https://github.com/thrust/thrust/issues/780 & https://github.com/thrust/thrust/issues/776

sdalton1:
Works for me, thanks!

thrust/detail/get_iterator_value.h
```diff
@@ -0,0 +1,49 @@
...
+namespace thrust {
+namespace detail {
+
+// get_iterator_value specialization on iterators
+// --------------------------------------------------
+// it is okay to dereference iterator in usual way
+template<typename DerivedPolicy, typename Iterator>
+__host__ __device__
+typename thrust::iterator_traits<Iterator>::value_type
+get_iterator_value(thrust::execution_policy<DerivedPolicy> &, Iterator it)
```
jaredhoberock: I'm not sure this implementation will work as intended because it ignores the execution policy. Shouldn't it do something like call `thrust::copy`?

thrust/detail/get_iterator_value.h
```diff
@@ -0,0 +1,49 @@
...
+// it may point to a location in the device memory. 
+// we use get_value(exec,pointer*) function
+// to perform a dereferencing consistent with the execution policy
+template<typename DerivedPolicy, typename Pointer>
+__host__ __device__
+typename thrust::detail::pointer_traits<Pointer*>::element_type 
+get_iterator_value(thrust::execution_policy<DerivedPolicy> &exec, Pointer* ptr)
+{
+  return get_value(derived_cast(exec),ptr);
+} // get_iterator_value(exec,Pointer*)
```
jaredhoberock: It looks like this implementation only works for raw pointers. What will happen with `thrust::device_ptr`?

thrust/detail/get_iterator_value.h
egaburov: Why wouldn't it, aren't true iterators supposed to be safe for usual derlerencing? If not, what is the interface to deference an iterator  with an execution policy? This passes all the tests.

thrust/detail/get_iterator_value.h
egaburov: Dereferencing device_ptr<T> is safe on the host. Upon dereferencing on the host data is copied from the device, and so it is treated as an iterator. This specialisation is meant for raw pointers only that might be unsafe to dereference on the host.

thrust/detail/get_iterator_value.h
jaredhoberock: I think that we intend for `get_iterator_value()` to be the interface to dereference an iterator with an execution policy. So, we have to build the implementation and ensure that it handles all cases correctly. I'm concerned that this implementation doesn't handle all cases correctly, namely, those cases where the user has provided a custom execution policy which is different from whatever is tagged in the iterator.

egaburov: I am not following. What would be an example of dereferencing an iterator with execution policy?

egaburov: To add to this: it will be a non-issue because nothing nothing changes for iterators: they are still  dereference via 'operator*', as it was earlier. If users were hitting bugs with such dereferencing, we would have already known. We need a counter example to show it is unsafe now and worked before this change.

jaredhoberock: Suppose I call `max_element(par.on(my_stream), vec.begin(), vec.end())` and look at the visual profiler timeline of my application. What sort of synchronization behavior would I observe with this implementation?

I believe I would see that all CUDA streams in my application synchronize at the call to `max_element` because the iterator dereference inside of `get_iterator_value()` will use the default stream instead of the stream contained inside of my execution policy.

Remember that in addition to preventing a crash due to dereferencing a raw device pointer on the host, we also have to ensure that the user's execution policy is used for this dereference. The one overload's use of `get_value()` ensures both of those things happen, but the other overload's plain iterator dereference does not.

I think that using `thrust::copy` will solve this problem.

egaburov: I understand. However, this fix doesn't change the previously existed behaviour, but fixes a bug when a raw pointer is passed to max_element.

That being said, replacing `*it` with 

      typename thrust::iterator_traits<Iterator>::value_type value;
      thrust::copy(exec, it,it+1, &value); 
      return value; 

makes the reproducer to die with

      what():  an illegal memory access was encountered
    Aborted (core dumped)

when used with transform_iterator and I wasn't able to quickly nail  down the issue. I agree the get_value_iterator needs to be enhanced to respect customer user policy, but it may take more time.

I suggest either accepting this fix or reverting https://github.com/thrust/thrust/pull/777 to make sure the user codes are not broken, and then filing a separate bug that requires get_iterator_value to respect user policy.

jaredhoberock:
Will merge this now and track the task of generalizing the implementation of `get_iterator_value()` elsewhere.

PR #790, opened 2016-05-04
1.8.3

egaburov:


egaburov:
We don't support pre 7.5 cuda with master branch. Closing PR.

PR #795, opened 2016-05-17
Avoid assignments and conversions to/from proxy references in block/c…

### LINKED ISSUE #793: Compilation error when sort is used with allocator and custom comparator.
Commit 0afb766f2bfdf0fe654f2b47de6add829df5d63f breaks sort when used with custom comparator. Here is the repro

```
$ cat r1.cu
#include <thrust/system/cuda/execution_policy.h>
#include <thrust/sort.h>

class Allocator
{
public:
  typedef char value_type;

  char *allocate(std::ptrdiff_t num_bytes) { return (char *)malloc(num_bytes); }
  void deallocate(char *ptr, size_t n) { free(ptr); }
};

struct cmp_int
{
  __host__ __device__
  bool operator()(const int &v0, const int &v1)
  {
    return v0 < v1;
  }
};

void thrust_sort_int2_by_x(int *vals, unsigned int num_elements,
                           Allocator *thrustAllocator, void *stream)
{
  thrust::sort(thrust::cuda::par(*thrustAllocator).on((cudaStream_t)stream),
               vals, vals + num_elements,
#ifdef PASS
              thrust::less<int>()
#else
               cmp_int()
#endif
      );
}

$ nvcc r1.cu -c -DPASS
$ nvcc r1.cu -c 
... errors
```

jaredhoberock:
…opy & stable_merge_sort

This avoids default-constructing execution policies in thrust::reference when calling
merge sort with a user-defined execution policy.

Fixes #793

### DIFF
--- a/thrust/system/cuda/detail/block/copy.h
+++ b/thrust/system/cuda/detail/block/copy.h
@@ -171,7 +171,7 @@ template<typename Context,
       first  += context.block_dimension(),
       result += context.block_dimension())
   {
-    *result = *first;
+    thrust::raw_reference_cast(*result) = thrust::raw_reference_cast(*first);
   } // end for
 
   return end_of_output;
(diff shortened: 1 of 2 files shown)


jaredhoberock:
@egaburov could you review this and see if it comprehensively fixes the issue? We might also want to add a test case for this in the unit tester.

egaburov:
Thanks Jared. I will test the fix and add the unit test.

egaburov:
Go ahead and merge. The unit tests and examples compiled and run successfully. I will create later a sperate PR with unit test for this functionality.

PR #808, opened 2016-06-04
1.8.3

egaburov:


egaburov:
Passed unit tests & examples in both c++11 & c++03. Merged manually

PR #811, opened 2016-06-08
fix warnings when compile with clang

egaburov:
No warnings are generated when thrust examples & tests are compiled with clang top of the tree. No additional regressions were introduced: what used to pass before, passes now, and same for fails.

### DIFF
--- a/thrust/system/cuda/detail/detail/set_operation.inl
+++ b/thrust/system/cuda/detail/detail/set_operation.inl
@@ -302,8 +302,8 @@ inline __device__
       // load the input into __shared__ storage
       __shared__ uninitialized_array<value_type, block_size * work_per_thread> s_input;
   
-      value_type *s_input_end1 = thrust::system::cuda::detail::block::copy_n(ctx, first1, subpartition_size.first,  s_input.begin());
-      value_type *s_input_end2 = thrust::system::cuda::detail::block::copy_n(ctx, first2, subpartition_size.second, s_input_end1);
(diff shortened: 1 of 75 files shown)


thrust/system/cuda/detail/detail/set_operation.inl
```diff
@@ -302,8 +302,8 @@ inline __device__
       // load the input into __shared__ storage
       __shared__ uninitialized_array<value_type, block_size * work_per_thread> s_input;
   
-      value_type *s_input_end1 = thrust::system::cuda::detail::block::copy_n(ctx, first1, subpartition_size.first,  s_input.begin());
-      value_type *s_input_end2 = thrust::system::cuda::detail::block::copy_n(ctx, first2, subpartition_size.second, s_input_end1);
+      value_type *s_input_end1    =    thrust::system::cuda::detail::block::copy_n(ctx, first1, subpartition_size.first,  s_input.begin());
+      /* value_type *s_input_end2 = */ thrust::system::cuda::detail::block::copy_n(ctx, first2, subpartition_size.second, s_input_end1);
```
jaredhoberock: Let's just eliminate `s_input_end2` and instead of introducing commented-out code. I'm afraid it will just raise questions over why this variable is commented out.

jaredhoberock:
Thanks Evghenii, it all looks fine except for the one bit about the commented-out code.

It looks like you want to merge into master and not 1.8.3, right?

egaburov:
Correct to master. 1.8.3 can't be compiled with clang today.

jaredhoberock:
Ah, that's right. Thanks.

egaburov:
Removed commented variables.

PR #825, opened 2016-08-16
 vector_base::clear is now implement with erase(begin(), end())

egaburov:


jaredhoberock:
Thanks Evghenni. I think switching to `.erase()` is definitely an improvement. I have one suggestion about the documentation of the enhancement. Let's avoid mentioning the implementation detail (`vector_base`) in the CHANGELOG summary and instead note that `.clear()` operations on vector types no longer require the element type to have a default constructor.

jaredhoberock:
Looks great, thanks!

PR #830, opened 2016-09-14
Sync attributes of reference::assign_from definitions with their declarations.

Artem-B:
Fixes compilation error with recent clang.


jaredhoberock:
Thanks Artem, it looks like a good change to me. It was probably just an omission that the attributes weren't applied to the definitions of these functions.

@egaburov: If you agree, I'll merge this into master.

Artem-B:
There was a bug in clang that caused H/D attributes to be inherited in some
cases.
assign_from() happened to trigger that bug so missing attributes went
unnoticed until now.

--Artem

egaburov:
Looks good to me.

PR #835, opened 2016-09-15
Mark functions that take the address of a __global__ function as host…

jlebar:
…-only.

clang is about to get pickier about disallowing references to things
from host+device code when it won't work on either host or device.

clang doesn't currently support launching kernels from the device side,
thus these __host__ __device functions that take a function pointer to
__global__ functions when __CUDA_RDC__ is not defined are no good.

Originally landed as 98b4e1680, reverted in 884d199e because the
condition was wrong. See
https://github.com/thrust/thrust/pull/831#pullrequestreview-61458

### DIFF
--- a/thrust/system/cuda/detail/bulk/detail/cuda_launcher/triple_chevron_launcher.hpp
+++ b/thrust/system/cuda/detail/bulk/detail/cuda_launcher/triple_chevron_launcher.hpp
@@ -74,7 +74,13 @@ struct triple_chevron_launcher_base<block_size,Function,true>
   __host__ __device__
   static global_function_pointer_t global_function_pointer()
   {
+    // Don't try to take the address of launch_by_value from the device side if
+    // we don't support launching kernels from __device__ functions.
+#if !defined(__CUDA_ARCH__) || (defined(__CUDACC_RDC__) && __CUDA_ARCH__ >= 350)
     return launch_by_value<block_size,Function>;
+#else
+    return NULL;
+#endif
   }
 };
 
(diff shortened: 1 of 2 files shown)


egaburov:
LGTM, but let me run a unit tester with -rdc=true to verify the correctness. Will report here.

egaburov:
All tests pass.

jaredhoberock:
Great, thanks!

jlebar:
Thank you, folks!

egaburov:
Bugger. This change must be reverted. I only tested it with -rdc=true -arch=sm_35 or higher. It will fail otherwise because kernel invocation, or address taking, must not be guarded by __CUDA_ARCH__.  I overlooked this aspect.

egaburov:
If it is guarded by __CUDA_ARCH__ device compilation will never specialize kernel. kaboom!

jaredhoberock:
Thanks for the analysis. It may be that this guard needs to be made clang-specific.

jlebar:
This may be easier if you guys write the patch?  I am happy to test it.

egaburov:
Wouldn't it be easier if clang would not disallow taking address of a kernel in a device code?

jlebar:
> Wouldn't it be easier if clang would not disallow taking address of a kernel in a device code?

Possibly, but clang emphasizes being a sound compiler.  :)

Taking the address of a function you cannot call isn't allowed in C++.  For example, you can't take the address of a private function you don't have access to.

Indeed taking addresses of functions from device code should probably be disallowed entirely, because indirect calls are not supported on the GPU.  We're not there yet, but this is a step in that direction.

gnzlbg:
Any progress on this?

jlebar:
I am happy to write another patch, but at this point I'm pretty confused about what the guard should be.

We checked in the code that makes this fail in clang a few days ago.

andrewcorrigan:
Can we please get this fixed as soon as possible?  Would changing the guard to only disable the code in question for `clang`, while leaving it alone for `nvcc`, be acceptable to everyone?
```
#if !(defined(__clang__) && defined(__CUDA__)) && (!defined(__CUDA_ARCH__) || (defined(__CUDACC_RDC__) && __CUDA_ARCH__ >= 350))
```

egaburov:
Since clang doesn't support Dynamic Parallelism, a simple guard should suffice:

```
#if !(defined(__clang__) && defined(__CUDA__))
```

 @jlebar Please submit PR and I will test it.

PR #838, opened 2016-09-16
Revert "Mark functions that take the address of a __global__ function as host…"

jaredhoberock:
Reverts thrust/thrust#835 due to bad interaction with `nvcc` 7.5.


egaburov:
It can be clang-specific like this:

instead of

```
  __host__ __device__
static global_function_pointer_t global_function_pointer() {
```

try this, assuming clang will never be able to do device-side launches:

```

#define THRUST_CLANG_RUNTIME_FUNCTION __host__ __device__
#if defined(__clang__) && defined(__CUDA__)
#define THRUST_CLANG_RUNTIME_FUNCTION ___host__
#endif 

THRSUT_CLANG_RUNTIME_FUNCTION
static global_function_pointer_t global_function_pointer() {

```

PR #840, opened 2016-10-10
attempt to resolve #835

andrewcorrigan:
This attempts to resolve Issue #835.

### DIFF
--- a/thrust/system/cuda/detail/bulk/detail/cuda_launcher/triple_chevron_launcher.hpp
+++ b/thrust/system/cuda/detail/bulk/detail/cuda_launcher/triple_chevron_launcher.hpp
@@ -74,7 +74,11 @@ struct triple_chevron_launcher_base<block_size,Function,true>
   __host__ __device__
   static global_function_pointer_t global_function_pointer()
   {
+#if !(defined(__clang__) && defined(__CUDA__))
     return launch_by_value<block_size,Function>;
+#else
+    return NULL;
+#endif
   }
 };
 
@@ -98,7 +102,11 @@ struct triple_chevron_launcher_base<block_size,Function,false>
   __host__ __device__
   static global_function_pointer_t global_function_pointer()
   {
+#if !(defined(__clang__) && defined(__CUDA__))
     return launch_by_pointer<block_size,Function>;
+#else
+    return NULL;
+#endif
   }
 };
 
--- a/thrust/system/cuda/detail/detail/launch_closure.inl
+++ b/thrust/system/cuda/detail/detail/launch_closure.inl
@@ -78,7 +78,11 @@ template<typename Closure,
   __host__ __device__
   static launch_function_t get_launch_function()
   {
+#if !(defined(__clang__) && defined(__CUDA__))
     return launch_closure_by_value<Closure>;
+#else
+    return NULL;
+#endif
   }
 
   template<typename DerivedPolicy, typename Size1, typename Size2, typename Size3>
@@ -119,7 +123,11 @@ template<typename Closure>
   __host__ __device__
   static launch_function_t get_launch_function(void)
   {
+#if !(defined(__clang__) && defined(__CUDA__))
     return launch_closure_by_pointer<Closure>;
+#else
+    return NULL;
+#endif
   }
 
   template<typename DerivedPolicy, typename Size1, typename Size2, typename Size3>


egaburov:
Both separable and whole program compilation unit tests pass. @jlebar does this fix issue with clang? @jaredhoberock please merge if it's okay.

jlebar:
wfm!

jaredhoberock:
I'm a little confused by the condition of the `#if` -- do these functions ever return anything other than `NULL` when the compiler is clang?

egaburov:
Nope, if compiler is clang the return is always NULL. Is this not the intention?

jaredhoberock:
I was expecting the function to return `NULL` only if the compiler is clang and if the function is called from a `__device__` function. In other words, in `__host__` code, the clang and nvcc behavior of the function would be identical. Is there something that prevents this behavior?

andrewcorrigan:
@jaredhoberock Did you see the last two comments at the bottom of #835.  Would you prefer the guard I posted there?

```
#if !(defined(__clang__) && defined(__CUDA__)) && (!defined(__CUDA_ARCH__) || (defined(__CUDACC_RDC__) && __CUDA_ARCH__ >= 350))
```

egaburov:
`nvcc` doesn't support guarding `__global__` functions with `__CUDA_ARCH__`  and it will result in UB. I suspect the same applies to clang: due to  `__CUDA_ARCH__` guard the device compilation pass will not see specialisation code, and will not compile `__global__` function (see repro below). @jlebar, can you confirm this?

As it seems the only remaining option is to either change function annotation, to `__host__`-only in the  clang mode, or permit clang to specialise `__global__` function in the device code.

Here  is simple repro that fails with both clang and nvcc w/o -rdc=true:

```
$ cat test.cu
$ cat test.cu
#include <cstdio>
#include <cassert>

template<class T>
__global__ void kernel(T *ptr)
{
  *ptr = 42;
}

__host__ void foo(int *ptr)
{
#if !(defined(__clang__) && defined(__CUDA__)) && (!defined(__CUDA_ARCH__) || (defined(__CUDACC_RDC__) && __CUDA_ARCH__ >= 350))
  kernel<<<1,1>>>(ptr);
#endif
}

__global__ void set_value(int*ptr)
{
  *ptr = 128;
}

int main()
{
  int *ptr;
  cudaMalloc(&ptr, sizeof(int));
  set_value<<<1,1>>>(ptr);
  int val;
  cudaMemcpy(&val, ptr, sizeof(int), cudaMemcpyDeviceToHost);
  printf("val_init= %d\n", val);
  foo(ptr);
  cudaMemcpy(&val, ptr, sizeof(int), cudaMemcpyDeviceToHost);
  printf("val_after= %d\n", val);
}
$ nvcc test.cu -arch=sm_35 -rdc=true && ./a.out
val_init= 128
val_after= 42
$ nvcc test.cu -arch=sm_35  && ./a.out
val_init= 128
$ clang test.cu  -L/usr/local/cuda/lib64 -lcudart --cuda-gpu-arch=sm_35 && ./a.out
val_init= 128
val_after= 128
```

If guard is commented, the code works as expected.

jaredhoberock:
Thanks for the explanation. Could we try the following:

1. for nvcc only, preserve the existing behavior for these functions
2. for clang only, return `NULL` in `__device__` code, and return the address of the `__global__` function in `__host__` code
3. Add a comment to these two sites explaining what's going on here

Is all that doable?

egaburov:
> 2. for clang only, return NULL in __device__ code, and return the address of the __global__ function in __host__ code

How do you see this implemented? It does not seem that we can use `__CUDA_ARCH__` guard to decide whether to return NULL  or ptr to `__global__` function in `__host__ __device__` function.  Similarly to `nvcc`, in clang both `__host__` and `__device__` code-paths must take pointer of `__global__` function for correct behaviour. @jlebar  can you confirm this?

I can only see an option to change function annotation. for clang only, annotate function as `__host__` but for nvcc `__host__ __device__`. This can open, however, other can of worms in clang mode.

jaredhoberock:
> Similarly to nvcc, in clang both __host__ and __device__ code-paths must take pointer of __global__ function for correct behaviour.

I see, thanks.

> I can only see an option to change function annotation. for clang only, annotate function as __host__ but for nvcc __host__ __device__.

It seems like modulating the annotation based on the compiler is the best choice here because what we're trying to do cannot be done with `__host__ __device__` functions in clang. Do we know what problems are introduced if we change the annotation to `__host__` for clang only?

egaburov:
I don't have a handle on what problems this may introduce; worse clang is silent when a `__host__` function is called from `__host__ __device__` function so we have no visibility if something goes wrong.

We can surely change annotation to `__host__` for clang-only, and `__host__ __device__` otherwise, and if nothing breaks in clang we can stick with it.

Alternatively, we can annotate it as `__host__` or `__host__ __device__` depending whether compiler support dynamic parallelism.  But then we have to test Thrust with nvcc and silence all calling `__host__` from `__host__ __device__` function warnings.

I'd say, ideally, allow clang to take address of  a `__global__` function in `__device__` mode, which appears to be less invasive. This will maintain backward compatibility with nvcc. There could more codes in the wild that do tricks like that.

jaredhoberock:
I agree it would be ideal if clang would allow taking the address of a `__global__` function in `__device__` code. Hopefully someday clang will enable that feature.

In the meantime, I suggest making these functions `__host__`-only for clang. That way, we can make Thrust work to the degree possible with present day clang, and it will preserve Thrust's existing behavior with nvcc. If some future clang enables taking the address of a `__global__` function in `__device__` code, we can eliminate this workaround.

So I think the remaining question is to figure out what problems are introduced if we make these functions `__host__`-only for clang only. Has anyone tested that option yet?

egaburov:
> I agree it would be ideal if clang would allow taking the address of a __global__ function in __device__ code. Hopefully someday clang will enable that feature.

Clang used to be able to do this. It is seems to be a  recent change that broke this functionality. The repro above with removed guard compiles and runs fine with clang as of Jun 20. I understand it does not compile with clang trunk anymore.

@jlebar, can this change be reverted in clang?

jlebar:
> worse clang is silent when a __host__ function is called from __host__ __device__ function so we have no visibility if something goes wrong.

This is actually an intentional design decision, motivated in no small part by projects such as Thrust.

Essentially every large template-metaprogramming-based CUDA project we've looked at calls H or D from HD.  nvcc emits a warning, causing builds of e.g. eigen (and I believe, Thrust) to spew thousands of diagnostics.  Calling D or H from HD may be "not allowed" (as the warning says), and I sort of agree from a moral standpoint that the world might be better if it actually were not allowed, but everybody does it and relies on it working.

In clang, we allow HD to call H, but, when compiling for host, it is a compile error to cause an HD function that calls D to be codegen'ed.  Similarly, when compiling for device, it's a compile error to cause an HD function that calls H to be codegen'ed.

AFAICT, nvcc tries to do roughly the same thing, but it's pretty easy to come up with examples where it does the wrong thing.  For example, nvcc from cuda 8 does not emit an error on the following code:

    template<typename T> __device__ T D() { return T(); }                                                                                                                                                           
    template<typename T> __host__ __device__ T HD() { return D<T>(); }                                                                                                                                          
    void host() { HD<int>(); }

In contrast, clang correctly identifies that we're calling a device function from host code and spits out an error.

> I'd say, ideally, allow clang to take address of a __global__ function in __device__ mode
> @jlebar, can this change be reverted in clang?

We've discussed this at length in previous PRs.  Again: This was an intentional change, and at the moment I don't believe it's a bug.  Specifically, I don't believe it's sound to allow taking the address of a G function when compiling in a mode where nested kernel calls are not allowed.  This follows from the rule in C++ that you cannot take the address of a function you cannot call.  This is a good thing, because otherwise you'd be able to dereference the function pointer and call the thing that you're not supposed to call.

To use a CUDA-specific example, H functions cannot call D functions.  So H functions should not be able to take the address of D functions.  If this were allowed, then the H functions could dereference said function pointer and then "run" the D function.  I have no idea what that should do.

For the same reason, I don't believe that, when compiling in a mode where nested kernel invocations are disallowed, it makes sense to allow taking the address of a G function.

egaburov:
Thanks Justin for details. In this case I agree with Jared, the path of least resistance is to annotate functions with `__host__` in clang-mode, and `__host__ __device__` otherwise.

egaburov:
Just to confirm I understand it correctly: clang is not worried that it breaks compatibility with nvcc existing feature set, and which may break existing CUDA codes out there that take address of `__global__`  function in `__device__` code.

jlebar:
Evghenii,

As someone who has worked for many months of my life on making a compiler that's compatible with nvcc, I assure you that I take this very seriously.  However, we occasionally have, after careful consideration, decided that clang should reject certain idioms that nvcc accepts.  In my last post to this thread, I gave an example of this.  There are plenty of others.  You have seen many of them go by in the PRs we've sent making Thrust compatible with clang.

We do not take the position that if nvcc accepts a block of code, clang has a bug if it rejects it.  Again, see the example in my previous post, where we believe that nvcc is wrong to accept the code I gave, and so we reject it.  One of the things users have told us that they like about clang is that it's stricter about some things at compile time -- this prevents them from shooting themselves in the foot.

As we've seen with Thrust, large projects often need some amount of porting in order to be compatible with clang.  In general, we've seen that this porting improves the overall quality of the project -- I hope that's the way you see the overall effect of the changes we've made to Thrust to make it compatible with clang.  To answer your question specifically, I am not worried about libraries that choose to support clang having to change code that uses this particular idiom, which I expect to be uncommon compared to the other issues they will have to address.

If it turns out that lots of user code relies on this particular idiom and it is a serious impediment to adoption of clang, I will be happy to revisit this decision.  I suspect that's not the case, but hey, I could be wrong.  We are still the new kid on the block.

Alternatively, I remain open to being convinced that clang is doing the wrong thing here -- but that argument needs to be made on the basis of the semantics of the CUDA language, not on the basis of what nvcc does.  So far, in the three threads where we have discussed this issue, nobody has attempted to explain why it is semantically correct to allow device code to take the address of a global function that it cannot call.  Nor has anyone attempted to explain what semantics this function pointer should have.

jaredhoberock:
I don't have a strong opinion about whether or not it makes sense for a function to take the address of another function that it cannot call. I would note however that `__host__` functions can take the address of `__global__` functions, but they can't really call `__global__` functions directly. The triple chevrons are really just some sugar wrapping something an API providing something similar to a remote procedure call. So, the function pointer has to be laundered through some other API to be usable, and you could imagine building something similar for `__device__` functions. For example, a `__device__` function could take the address of a `__global__` function and communicate with some host thread set aside for launching kernels as directed by a CUDA thread.

What concerns me about the problem discussed in this thread is that there's apparently not a workaround based on `__CUDA_ARCH__` to guard `__host__ __device__` functions from doing illegal things. If unguarded `__host__ __device__` functions aren't allowed to take the address of `__global__` functions, that's fine. But it seems like it ought to be legal to do so if we guard on `__CUDA_ARCH__` to ensure that it only occurs when executed in the `__host__` part of the function.

Anyway, @andrewcorrigan: can you try implementing the workaround suggested [here](https://github.com/thrust/thrust/pull/840#issuecomment-252764158)? If everyone agrees it works without introducing any problems, we'll be ready to merge.

andrewcorrigan:
@jaredhoberock Your wish is my command. Please see #841

jlebar:
> I would note however that __host__ functions can take the address of __global__ functions, but 
> they can't really call __global__ functions directly. The triple chevrons are really just some sugar 
> wrapping an API providing something similar to a remote procedure call.

Yes, this is a problem.  I think it would make sense to say that you can't take the address of a `__global__` function, or alternatively, if we have to allow it, we could say that the fact that it's a kernel is part of the function's signature, such that it's illegal to call it without triple angle brackets.

Maybe that ship has sailed, but it's something I've thought some about.

> For example, a __device__ function could take the address of a __global__ function and 
> communicate with some host thread set aside for launching kernels as directed by CUDA
> threads.

I suppose...  You wouldn't actually be able to use this directly, though, because the address of the global function on the device isn't going to be the same as the address of said function on the host.  You'd have to maintain a mapping from device addresses to host addresses on the host, and at that point...use an enum?

> If unguarded __host__ __device__ functions aren't allowed to take the address of __global__ functions, that's fine. But it seems like it ought to be legal to do so if we guard on __CUDA_ARCH__ to ensure that it only occurs when executed in the __host__ part of the function.

So the canonical way of solving this in clang, I think, is to write a `__host__` and a `__device__` overload of the function in question.  That is, clang lets you do

    int f() { return 0; }
    __device__ int f() { return 1; }
    __host__ __device__ int hd() { f(); }

This will return 0 in host mode and 1 in device mode.

If everything is templated nvcc may accept this, I am not sure.

If that doesn't work, you could try moving the reference to the kernel function into a `__host__` function and calling *that* from your HD function.

    __host__ kernel_f_ptr get_ptr_host() {
      return kernel<int>;
    }
 
    __host__ __device__ kernel_f_ptr get_ptr()
    {
#if !defined(__CUDA_ARCH__)
      return get_ptr_host();
#else
      return NULL;
#endif
    }

Problem is this doesn't work if get_ptr_host is itself a templated function that we don't observe being instantiated when we compile for device.  But I'm sure you're clever enough to come up with a way to make that happen...

jlebar:
Here's a of the code above, if it helps.  This works fine in clang, because we see the kernel being instantiated in the free host function.  https://gist.github.com/3a3024946508d286f3af47e2705b651e

egaburov:
Code works with nvcc as well because kernel is instantiated in non-templated host function. It appears clang has the same behaviour as nvcc when it comes to guarding kernels instantiation with `__CUDA_ARCH__`. The big advantage of the clang is overload on `__host__` and `__device__` attribute, something that nvcc lack, and this allows to build a proper solution.

If I may suggest, can we change https://github.com/thrust/thrust/pull/841 to take advantage of this.

In a sense we can have the following

```
 __host__ __device__
static global_function_pointer_t global_function_pointer()		
{		    
  return launch_by_value<block_size,Function>;		    
}		    
```

to

```
#if CLANG_DEVICE_COMPILER
 __host__ 
static global_function_pointer_t global_function_pointer()		
{		    
  return launch_by_value<block_size,Function>;		    
}
__device__ 
static global_function_pointer_t global_function_pointer()		
{		    
  return NULL;
}
#else
 __host__ __device__
static global_function_pointer_t global_function_pointer()		
{		    
  return launch_by_value<block_size,Function>;		    
}
#endif

andrewcorrigan:
@egaburov   Looks like I read your mind.

jaredhoberock:
Thanks for the reminder about overloading on `__host__` and `__device__`. I agree that is the best way to handle this problem and it looks like @andrewcorrigan already has a proof of concept. If tests pass, then let's merge pull request #842.

PR #841, opened 2016-10-11
__device__ functions cannot take the address of __global__ functions …

andrewcorrigan:
…in clang

How about now?  @jlebar @jaredhoberock @egaburov


andrewcorrigan:
Nevermind

PR #842, opened 2016-10-11
use overloading with clang to avoid taking a reference to a __global_…

andrewcorrigan:
…_ function in __device__ code

"So the canonical way of solving this in clang, I think, is to write a __host__ and a __device__ overload of the function in question." https://github.com/thrust/thrust/pull/840#issuecomment-252785835 

Since Clang allows overloading, how about making use of that?   @jaredhoberock @egaburov @jlebar


egaburov:
You beat me on the idea :) Having came to same idea independently  probably tells it is not such a bad one after all.

I'd suggest one slight improvement though. It would be good to fix macro in https://github.com/thrust/thrust/blob/master/thrust/detail/config/compiler.h#L77 to set `THRUST_DEVICE_COMPILER` to `THRUST_DEVICE_COMPILER_CLANG`

It appears we use 

```
#if (defined(__clang__) && defined(__CUDA__))
```

in many places in Thrust, and it would be  a nightmare to change if the signature of clang device compiler will change. 

@andrewcorrigan can you please adjust https://github.com/thrust/thrust/blob/master/thrust/detail/config/compiler.h#L77  correspondingly, so that we can just type

```
#if THRUST_DEVICE_COMPILER  == THRUST_DEVICE_COPMILER_CLANG
<clang code>
#else
<default code>
#endif
```

Thanks

andrewcorrigan:
working on it.

egaburov:
The change looks good to me. Unit tests pass with nvcc in both separable and whole program compilation mode.

jlebar:
> it would be a nightmare to change if the signature of clang device compiler will change.

We would never *dream* of such a thing.  :)

I'm happy we were able to settle on something we're all happy with.

thrust/system/cuda/detail/bulk/detail/cuda_launcher/triple_chevron_launcher.hpp
```diff
@@ -71,11 +71,25 @@ struct triple_chevron_launcher_base<block_size,Function,true>
...
+#if THRUST_DEVICE_COMPILER == THRUST_DEVICE_COMPILER_CLANG
+  __host__
+  static global_function_pointer_t global_function_pointer()
+  {
+    return launch_by_value<block_size,Function>;
+  }
+
+  __device__
+  static global_function_pointer_t global_function_pointer()
+  {
```
egaburov: Can we add a comment saying that clang doesn't support dynamic parallelism, and add assert(0) or something like this to terminate execution should this code path be ever taken at runtime?

andrewcorrigan: Is `assert(0)` ok in `__device__` code?  How about `bulk::detail::terminate()` instead?   If you want analogous changes in `thrust/system/cuda/detail/detail/launch_closure.inl`, can that call directly into `bulk::detail::terminate()` too?

egaburov: assert(0) works in nvcc, don't know about clang. bulk::detail::terminate() is better I think.

andrewcorrigan: I used `bulk::detail::terminate()` in bulk and `assert(0)` otherwise.  Anything else?

thrust/detail/config/compiler.h
```diff
@@ -71,10 +71,10 @@
 #define THRUST_DEVICE_COMPILER THRUST_DEVICE_COMPILER_GCC
 #elif THRUST_HOST_COMPILER == THRUST_HOST_COMPILER_CLANG
 // CUDA-capable clang should behave similar to NVCC.
-#if defined(__CUDA__)
-#define THRUST_DEVICE_COMPILER THRUST_DEVICE_COMPILER_NVCC
-#else
+#if defined(__CUDA__) && defined(__clang__)
```
Artem-B: `#if defined __CUDACC__` on line 66 would still set compiler to THRUST_DEVICE_COMPILER_NVCC and we'll never make it here.

Artem-B: Furthermore, if THRUST_HOST_COMPILER == THRUST_HOST_COMPILER_CLANG, then it will trigger static asserts in number of places that expect compiler for CUDA code to be NVCC.
For instance there's a static assert in thrust/system/cuda/detail/for_each.inl:109

PR #843, opened 2016-10-11
Use old-style check for clang

Artem-B:
.. until the rest of clang sources are ready for THRUST_DEVICE_COMPILER_CLANG


egaburov:
so be it :) Thanks Artem

jaredhoberock:
Thanks Artem. Just so we understand, the current definition of `THRUST_DEVICE_COMPILER` is still broken, correct?

Artem-B:
On Tue, Oct 11, 2016 at 4:59 PM, Jared Hoberock <notifications@github.com>
wrote:

Thanks Artem. Just so we understand, the current definition of
> THRUST_DEVICE_COMPILER is still broken, correct?
>
It is broken in the sanse that it does not do what it was intended to do.
We neither set it to _CLANG, nor handle it correctly if it were set that
way.
Fortunately former saves us from later, so #842 by itself is more-or-less a
no-op.

PR #844, opened 2016-10-12
Revert the clang-specific kernel launching code

jlebar:
Revert shenanigans for launching kernels with clang.  …
The code being reverted compiles, but doesn't actually work -- in fact,
we end up generating no code on the device at all!

The problem is that `triple_chevron_launcher_base` is a templated class,
and so we only instantiate functions that are actually called.  When
compiling for device, we would never instantiate the `__host__` function,
so we never observed the kernel being referenced, so we never generated
any code for it.

clang was actually wrong to reject this code.  If the launch functions
were `__device__`, clang would have been right.  But being `__host__`
`__device__`, clang should wait to emit an error until it sees the HD
function being *used* on the device side.  If the function is never used
on the device side, it's fine to do any host-only things from the HD
function.

As of r283963 [1], clang now does the right thing.

[1] llvm-mirror/clang@f3d02c1

### DIFF
--- a/thrust/system/cuda/detail/bulk/detail/cuda_launcher/triple_chevron_launcher.hpp
+++ b/thrust/system/cuda/detail/bulk/detail/cuda_launcher/triple_chevron_launcher.hpp
@@ -71,27 +71,11 @@ struct triple_chevron_launcher_base<block_size,Function,true>
 {
   typedef void (*global_function_pointer_t)(Function);
 
-#if defined(__CUDA__) && defined(__clang__)
-  __host__
-  static global_function_pointer_t global_function_pointer()
-  {
-    return launch_by_value<block_size,Function>;
-  }
-
-  __device__
-  static global_function_pointer_t global_function_pointer()
-  {
-    bulk::detail::terminate(); //  clang doesn't support dynamic parallelism
-
-    return NULL;
-  }
-#else
   __host__ __device__
   static global_function_pointer_t global_function_pointer()
   {
     return launch_by_value<block_size,Function>;
   }
-#endif
 };
 
 
@@ -111,27 +95,11 @@ struct triple_chevron_launcher_base<block_size,Function,false>
 {
   typedef void (*global_function_pointer_t)(const Function*);
 
-#if defined (__CUDA__) && defined(__clang__)
-  __host__
-  static global_function_pointer_t global_function_pointer()
-  {
-    return launch_by_pointer<block_size,Function>;
-  }
-
-  __device__
-  static global_function_pointer_t global_function_pointer()
-  {
-    bulk::detail::terminate(); //  clang doesn't support dynamic parallelism
-    
-    return NULL;
-  }
-#else
   __host__ __device__
   static global_function_pointer_t global_function_pointer()
   {
     return launch_by_pointer<block_size,Function>;
   }
-#endif
 };
 
 
@@ -241,3 +209,4 @@ class triple_chevron_launcher<block_size_,Function,false> : protected triple_che
 } // end detail
 } // end bul
 BULK_NAMESPACE_SUFFIX
+
--- a/thrust/system/cuda/detail/detail/launch_closure.inl
+++ b/thrust/system/cuda/detail/detail/launch_closure.inl
@@ -75,27 +75,11 @@ template<typename Closure,
 {
   typedef void (*launch_function_t)(Closure); 
  
-#if defined(__CUDA__) && defined(__clang__)
-  __host__
-  static launch_function_t get_launch_function()
-  {
-    return launch_closure_by_value<Closure>;
-  }
-
-  __device__
-  static launch_function_t get_launch_function()
-  {
-    assert(0); // clang doesn't support dynamic parallelism
- 
-    return NULL;
-  }
-#else
   __host__ __device__
   static launch_function_t get_launch_function()
   {
     return launch_closure_by_value<Closure>;
   }
-#endif
 
   template<typename DerivedPolicy, typename Size1, typename Size2, typename Size3>
   __host__ __device__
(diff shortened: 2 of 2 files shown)


jlebar:
@andrewcorrigan, @gnzlbg, @egaburov, @jaredhoberock -- There *was* a bug in clang, just not the one we all thought it was.  See the commit message, I hope that explains things.

Now, if you actually *want* to touch this HD function from the device side...then I'm back to square one.  But for now this works for me, passes tests locally.

egaburov:
Does it mean that `__host__` & `__device__` overload at the language level is to the effect is no different than an approach of guarding with `__CUDA_ARCH__` preprocessor macro in `__host__ __device__` code?

egaburov:
>  If the launch functions were `__device__`, clang would have been right. But being `__host__
__device__`, clang should wait to emit an error until it sees the HD function being used on the device side.

I'd make an argument here that it is legal to take address of a `__global__` function  in the `__device__` code as well, and report an error only if it is used to launch a kernel. 

I'd like to stress that it is legal to take address of an entry function in the ptx, whether dynamic parallelism is enabled or not. I can take the address of the `__global__` function  in CUDA, copy it to the host, and do interesting things with it.

jlebar:
> Does it mean that __host__ & __device__ overload at the language level to the effect is no different than an approach of guarding with __CUDA_ARCH__ preprocessor macro in __host__ __device__ function?

No.  If the functions had not been templated, things would have worked as I expected.

"The problem is that triple_chevron_launcher_base is a templated class, and so we only instantiate functions that are actually called. When compiling for device, we would never instantiate the `__host__` function, so we never observed the kernel being referenced, so we never generated any code for it."

> I'd make an argument here that it is legal to take address of a __global__ function in the __device__ code as well, and report an error only if it is used to launch a kernel.

That's not statically-analyzable.  We can't tell what you do with the pointer after you take the address of the function, so in particular we cannot tell if you call it.  It is undecidable.

egaburov:
> That's not statically-analyzable. We can't tell what you do with the pointer after you take the address of the function, so in particular we cannot tell if you call it. It is undecidable.

I don't follow, can't clang see that a function pointer is called with <<<,>>> notations in the device code and report an error then. As far as I don't launch a kernel with a function pointer, it seems to me it should be legal to take it address in a device function, if we make it legal in host device one.

jlebar:
> clang see that a function pointer is called with <<<,>>> notations in the device code and report an error then

Oh, you're just saying, any function pointer should not be able to be called by `<<<,>>>`.  I see.

As I've said a few times now, for me this comes back to the notion that in C++ you cannot take the address of a function that you cannot call.  I've also already discussed elsewhere why I think it's unlikely that you'd be able to do something useful with said address on the host side.  

Nevertheless, if you can find existing OSS CUDA code that, when running in a mode where nested kernels are not allowed

* takes the address of a kernel from device code and then does something interesting with it on the host side, and

* could not easily be rewritten to a different idiom,

then I would seriously consider accepting a patch to clang to make this change.  (Because none of my customers want this, it is very unlikely that anyone from Google will spend the time actually writing said patch.)

Honestly it might be easier and less controversial just to write the patch allowing nested kernel calls.  I would gladly take that patch.

egaburov:
> Oh, you're just saying, any function pointer should not be able to be called by <<<,>>>. I see.

In device code, yes. If there is no dynamic parallelism, no function pointer must be called with <<<,>>>. With dynamic parallelism, only `__global__` function pointer can be called with <<<,>>>, but this cannot be enforced at the type level.

Clang surely can do what it sees fit, but what I am rather concerned here is that clang imposes  seemingly artificial constraints on the device code, but does nothing of this sort in the host code, viz. the following:

In the host code one can take pointer of a `__global__` function, and call it. This will however result in UB. Clang can't catch this during semantic analysis (or can it?).  A  `__global__` function pointer must be invoked via <<<,>>>, but since `__global__` is not a part of the  function signature this cannot be enforced at type level; fwiw, reverse is also true, calling <<<,,>>> on host function pointer can't be caught and will result in UB.

And yet clang decides to hammer this problem  in the device code by disallowing taking pointer of a `__global__` function, even though this is perfectly legal. It sounds only logical that then clang should also disallow taking pointer to `__global__` function in the host code, because such function pointer is not callable  in C++ sense, but this will break a lot of codes and has no easy fix.

It seems to me a pragmatic approach would be to allow taking pointer of `__global__` function in both device and host code, because it is perfectly legal, compiles, and works. But the moment a function pointer is attempted to be invoked with <<<,>>> notation, report a compilation error in device code. If such function pointer is just called, well, lights out, UB in both host and device code... 

Makes sense?

jlebar:
Evghenii, the ramifications of taking the address of a global in host code have already been discussed previously.  I understand that you are frustrated and think clang is doing the wrong thing here, and I apologize.  As said above, I think your position is defensible, and I would be willing to take a patch changing clang's behavior, if we could motivate the additional complexity with some real code that relies on it.

However, currently this feature is not important to my customers, and at the moment we've found no other code that relies on it, so it's not a priority for me.  Maybe that will change tomorrow!  Or maybe I'll get to it during a dull period.  I hope that makes sense.  Definitely if you can find an existing use-case, that would make it more likely that we would do this.

I believe you work for nvidia -- sorry if my sleuthing turned up the wrong result.  If you all are interested in collaborating more closely with us on clang, I think that would be a win-win, and it would allow me to focus some of my time on issues like this one that are important to nvidia.  A good place to start might be https://llvm.org/bugs/show_bug.cgi?id=27738, which is an apparent ptxas miscompile in Thrust.  So far all we've heard is that this is a known bug in ptxas and there's no workaround.

In any case, can we please merge this PR so that our customers are unbroken?

egaburov:
Thanks for tee details. No worries, I want to understand Google view on CUDA support in clang, no frustrations and agree to disagree ^_^

PR #847, opened 2016-10-16
Added the code for Deshpande et. al's paper

dracarys983:
Hello!
So this is the pull request that Jared told me to create in this topic I posted on thrust-users mailing list : https://groups.google.com/forum/#!topic/thrust-users/jXGUPy3cN2M

I have not adhered to any guidelines in my code as of now, so it might need some changes.


jaredhoberock:
Thanks for the submission!

The implementation is a bit long for an example program (several hundred lines of code), and seems to require some CUDA kernels. I worry that the current incarnation may be too complex for an effective pedagogical example program. It may make for a better blog post which could go into more detail explaining the complexities of the implementation.

Do you see any avenues for simplification and portability?

dracarys983:
Hello Jared!
I believe there is not much possibility for simplification in this code. I will try to simplify it as much as possible though and add some important comments. We can check then if it's better to have a blog post instead.

Does that sound good?

jaredhoberock:
Sounds good to me.

brycelelbach:
The license on this example is an issue. The code would need to be relicensed under the Apache license.

brycelelbach:
Ping re licensing.

dracarys983:
Is the code alright to be added as is?
I have not been able to make any further reductions or convert to a smaller example for pedagogical purposes.

If the code can be added as is, I can do the re licensing after talking to the author.

brycelelbach:
@jaredhoberock mind giving this another peak?

brycelelbach:

allisonvacanti:
@brycelelbach @jaredhoberock @dracarys983 It's been over a year since there was any activity on this PR and the diff has gotten unruly. I'm going to close this PR -- please rebase on current master and resubmit if you'd like.

Also -- re: licensing, we cannot merge code that is under an incompatible license. We'd need that addressed before we can consider integrating this into thrust.

PR #848, opened 2016-10-20
Sync few more attribute mismatches exposed by clang.

Artem-B:


jaredhoberock:
Looks good to me. Do you agree, @egaburov?

egaburov:
+1. Will merge

PR #851, opened 2016-10-24
fixes error introduced in #842

### LINKED ISSUE #845: compile error: unknown type name '__host__' from clang++
The following code does not compile after the changes in THRUST_DEVICE_COMPILER_CLANG detection. (It used to work correctly before)
```
 #define THRUST_DEVICE_SYSTEM THRUST_DEVICE_SYSTEM_CPP
 #include <thrust/system/cpp/vector.h>
 int main()
 {
    thrust::cpp::vector<float> v(44);
 }
```
The system is linux. We tested with clang versions 3.6 and 3.8.1.

andrewcorrigan:
I think that this should fix #845. I'm sorry for introducing this regression.

### DIFF
--- a/thrust/detail/config/compiler.h
+++ b/thrust/detail/config/compiler.h
@@ -73,7 +73,7 @@
 // CUDA-capable clang should behave similar to NVCC.
 #if defined(__CUDA__) && defined(__clang__)
 #define THRUST_DEVICE_COMPILER THRUST_DEVICE_COMPILER_CLANG
-#else
+#elif defined(__CUDA__)
 #define THRUST_DEVICE_COMPILER THRUST_DEVICE_COMPILER_NVCC
 #endif
 #else


jaredhoberock:
Thanks, Andrew. It looks like this change rolls back that line to what it was before Artem-B's recent change to `compiler.h`, is that right?

egaburov:
We should revert this file to the state it had before PR #842? It appears we do not fully understand interaction of different macros, and until we do it we should revert to what used to work.

andrewcorrigan:
I didn't roll it back completely, but I think this fixes the error reported in #845 and is more correct than the check was before.  Basically, the cumulative changed from before my previous PR is that this:
```
#if defined(__CUDA__)
#define THRUST_DEVICE_COMPILER THRUST_DEVICE_COMPILER_NVCC
#else
#define THRUST_DEVICE_COMPILER THRUST_DEVICE_COMPILER_CLANG
#endif
```
which doesn't make sense but is innocuous, changes now to:
```
#if defined(__CUDA__) && defined(__clang__)
#define THRUST_DEVICE_COMPILER THRUST_DEVICE_COMPILER_CLANG
#elif defined(__CUDA__)
#define THRUST_DEVICE_COMPILER THRUST_DEVICE_COMPILER_NVCC
#endif
```
Which is correct, but ineffective, as pointed out by Artem.

The error message reported in #845, I believe is due to my change which incorrectly made the above:
```
#if defined(__CUDA__) && defined(__clang__)
#define THRUST_DEVICE_COMPILER THRUST_DEVICE_COMPILER_CLANG
#else
#define THRUST_DEVICE_COMPILER THRUST_DEVICE_COMPILER_NVCC
#endif
```
which failed to actually check for `__CUDA__` before setting `THRUST_DEVICE_COMPILER_NVCC`.

andrewcorrigan:
@egaburov  Whatever you guys prefer. I can submit a different PR with the change just rolled back.

Another option: I'd be happy to get `THRUST_DEVICE_COMPILER_CLANG` actually working too, and deal with all the rest of the code that only allows for `THRUST_DEVICE_COMPILER_NVCC`.

jaredhoberock:
Agreed with @egaburov - let's roll it back to the state it was before and then try to fix the issue separately.

andrewcorrigan:
Please see #853.

PR #853, opened 2016-10-24
rollback #842

### LINKED ISSUE #845: compile error: unknown type name '__host__' from clang++

andrewcorrigan:
If you guys prefer this over #851, this should also fix #845.


jaredhoberock:
It looks like this rolls things back to Artem's last change to this file: https://github.com/thrust/thrust/blob/33a4e1fae8b8bb380cd8fb0f94fe271d61c236cf/thrust/detail/config/compiler.h#L73

Looks OK to me, we can merge if @egaburov agrees.

egaburov:
LGTM

PR #857, opened 2016-10-26
thrust_exec_check_disable

andrewcorrigan:
Just like #570.

### DIFF
--- a/thrust/detail/allocator/allocator_traits.inl
+++ b/thrust/detail/allocator/allocator_traits.inl
@@ -100,6 +100,7 @@ template<typename Alloc, typename T, typename Arg1>
     : has_member_construct2_impl<Alloc, void(T*,const Arg1 &)>
 {};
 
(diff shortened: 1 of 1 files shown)


jaredhoberock:
Looks fine to me. The annotation is required because `.construct()` may be a `__host__` function.

egaburov:
Looks OK. One followup question though: is it because `.construct()` may come from host STL allocators, that are unannotated with `__host__ __device__`?

andrewcorrigan:
I didn't pay too much attention to the details of the error messages in this instance, but last time [1] it was:

> storing a type with `__device__ __host__` constructor in a `thrust::host_vector`, which invokes the constructor from `__host__` code.

[1] https://github.com/thrust/thrust/issues/556#issuecomment-59660454

jaredhoberock:
@egaburov : Yes, that's correct.

PR #878, opened 2017-04-18
Optimization min/max_element when using basic types

fran6co:
get_iterator_value triggers a sync in CUDA


brycelelbach:
NVBug 2157884. Running tests now.

PR #886, opened 2017-08-01
Fixed minor documentation error with thrust::minus

tim510:
The output that the documentation stated should result from the example code was incorrect, and was the output that would have resulted if the arguments to thrust::transform were inverted.


brycelelbach:
Thanks.

PR #899, opened 2017-12-13
Update pinned_allocator.h

TurboLed:
operator== and operator!= should be const. This is causing compilation issues on VS 2015 when assigning two std::vector<T, pinned_allocator<T>>


brycelelbach:
Done.

brycelelbach:
NVBug 2157926

PR #887, opened 2017-08-03
fixed a typo

tschmidt23:

PR #914, opened 2018-05-29
Fixed a typo in the documentation.

brycelelbach:
This is #887 with an amended commit message.


brycelelbach:
Merged into CUDA Thrust as well.

PR #957, opened 2019-02-28
Modified thrust examples to return explicit test results

brycelelbach:
Continuation of #646.


brycelelbach:

PR #966, opened 2019-04-01
Fix C4244 from MSVC in sort.

zhihaoy:


zhihaoy:
The issue goes away in MSVC 19.16.27030.1, should I leave this P/R open?

brycelelbach:
@zhihaoy, hrm, your call. This probably isn't the right fix; instead of casting, we should change the type of the variable. @griwes, thoughts?

zhihaoy:
It's a safe fix because "log2 round to integer" can't overflow any type.  Raising `num_passes` to `Size` looks pedantic... I will leave it open for a while because someone else seems to encounter the same issue https://devtalk.nvidia.com/default/topic/1064002/b/t/post/5387873/

allisonvacanti:
This patch looks reasonable to me -- we wouldn't be running more than 2^32 passes so just casting to an `int` should be ok. @brycelelbach @griwes if you have any objections to this let me know, but I'll start moving towards integrating this in case folks are still using older MSVC 2017 builds.

allisonvacanti:
Started internal CI: 28483968

allisonvacanti:
Tests pass -- I'll merge this soon.

PR #1005, opened 2019-07-19
thrust::system::cuda::allocator constructor should be host only

### LINKED ISSUE #972: stateless_resource_allocator generates warnings about using __host__ function from __host__ __device__
When building VTK-m ( https://gitlab.kitware.com/vtk/vtk-m/ ) with CUDA 10.1 we are seeing thrust generate warnings the following warning:
```
... include/thrust/mr/allocator.h(106): warning #2912-D: calling a __host__ function("thrust::mr::stateless_resource_allocator<long long,  ::thrust::system::cuda::detail::cuda_memory_resource<& ::cudaMalloc, & ::cudaFree,  ::thrust::cuda_cub::pointer<void> > > ::stateless_resource_allocator") from a __host__ __device__ function("thrust::mr::stateless_resource_allocator<long long,  ::thrust::system::cuda::detail::cuda_memory_resource<& ::cudaMalloc, & ::cudaFree,  ::thrust::cuda_cub::pointer<void> > > ::stateless_resource_allocator [subobject]") is not allowed
```

robertmaynard:
Fixes #972.
The constructor of this allocator has no need to be device constructible.

### DIFF
--- a/thrust/system/cuda/memory.h
+++ b/thrust/system/cuda/memory.h
@@ -100,7 +100,7 @@ public:
 
   /*! No-argument constructor has no effect.
    */
-  __host__ __device__
+  __host__
   inline allocator() {}
 
   /*! Copy constructor has no effect.


robertmaynard:
@griwes I finally remembered to make a merge request after we talked about how to fix this warning, as seen by VTK-m.

brycelelbach:
There's a bunch of other constructors that are `__host__ __device__`. The right change is either to make them all `__host__`, or to make the underlying facilities (`stateless_resource_allocator`) `__host__ __device__`.

We need to give this some more thought, as @griwes said.

brycelelbach:
@griwes makes the good point that the default constructor actually does host work, while the copy constructors and the destructors don't. So perhaps this is the right change.

brycelelbach:
@griwes can you code review this?

We need to decide what to do here.

Running internal CI on this.

allisonvacanti:
We collectively spent at least two days trying to track down this warning while this PR just sat here...

D'oh.🤦

Already fixed in d43f28542ed9a9ce1d61e959612a926ee1d48698. Closing.

PR #1010, opened 2019-08-09
use c++11 auto for iterator types

rongou:
Seems to make the examples cleaner. :)

### DIFF
--- a/examples/arbitrary_transformation.cu
+++ b/examples/arbitrary_transformation.cu
@@ -59,9 +59,9 @@ int main(void)
     A[4] = 2;  B[4] = 8;  C[4] = 3; 
 
     // apply the transformation
-    thrust::for_each(thrust::make_zip_iterator(thrust::make_tuple(A.begin(), B.begin(), C.begin(), D.begin())),
-                     thrust::make_zip_iterator(thrust::make_tuple(A.end(),   B.end(),   C.end(),   D.end())),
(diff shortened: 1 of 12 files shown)


brycelelbach:
Running internal CI, but that has coverage gaps for C++11.

@griwes and @jaredhoberock to code review.

brycelelbach:
Doing this without a version check is problematic, as we still have to support C++03. I'm not sure that we can take this as-is.

jaredhoberock:
Once the example programs can use `auto` without introducing other problems, then I think that's a good idea.

The problem I see is that we'd need to communicate to the user somehow that their compiler needs to be set to a C++ dialect that includes `auto`. The example programs are intended to be pedagogical, and understandable by novice users that might not even be that familiar with C++. I worry that adding that additional an detail concerning dialect would present an obstacle to new learners.

brycelelbach:
So, unfortunately, I think until we can move all of Thrust to be C++11 only, this will only complicate the examples; either we make all of these examples C++11 only, or we add `#ifdef`s to the examples, which makes them harder to read.

So, for now, let's shelve this.

brycelelbach:
But thanks so much for the PR - we'll revisit this when we move to C++11 only.

henryiii:
I'm still not convinced there are that many users eagerly upgrading to the latest cutting edge Thrust that are also still stuck on C++98...

PR #1020, opened 2019-10-10
Clear CUDA global error state after CUDA API calls

brycelelbach:
After making a CUDA API call, always clear the global CUDA error state by calling `cudaGetLastError`. Otherwise, if the CUDA API call is followed directly by a kernel launch, checking for a synchronous error during the kernel launch by calling `cudaGetLastError` may potentially return the error code from the CUDA API call. This type of error leakage is very subtle and difficult to trace.

This is a fix for Bug 2720132.

### DIFF
--- a/doc/branching.md
+++ b/doc/branching.md
@@ -13,6 +13,7 @@ versions don't directly map to any CUDA Toolkit version.
 
 | Thrust version    | CUDA version  |
 | ----------------- | ------------- |
+| 1.9.6             | 10.1 Update 2 |
 | 1.9.5             | 10.1 Update 1 |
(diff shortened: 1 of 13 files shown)


griwes:
So... isn't there a danger of silently ignoring an asynchronous error if we do this this way? Shouldn't we actually be checking the return value of the `cudaGetLastError` call and reporting unexpected errors in one way or another?

brycelelbach:
No. Asynchronous errors are sticky; they cannot be cleared.

brycelelbach:
This passed internal CI. Accepting.

PR #1021, opened 2019-10-10
Don't call the destruction kernel for an empty vector

brycelelbach:
When destroying a Thrust vector, we should avoid calling `destroy` if the vector is empty. For a `device_vector`, destroy calls into parallel algorithm machinery, which doesn't currently early exit (or early exit enough) for empty ranges.

This is a fix for Bug 2720132.


brycelelbach:
Internal CI has passed. Merging this.

PR #1026, opened 2019-10-21
CMake: Use correct MSVC version

bjude:
The cmake VERSION_LESS comparison expects versions to be formatted as major.minor.patch.tweak, the current MSVC version is 19.23.28105.4.
The current comparison checking agains 1900.0.0.0 instead of 19.0.0.0, this PR corrects that


brycelelbach:
Again, looks like an oversight.

Running internal CI.

@griwes to review.

brycelelbach:
Passed internal CI (no surprises, it doesn't actually use CMake). Accepting.

PR #1027, opened 2019-10-21
Fix tuple_for_each unused variable macro and tests

bjude:
The incorrect macro is used for removing the unused variable warning so this PR corrects that and adds tests for the other 2 tuple algorithms

### DIFF
--- a/testing/tuple_algorithms.cu
+++ b/testing/tuple_algorithms.cu
@@ -16,6 +16,16 @@ struct custom_square
   }
 };
 
+void test_tuple_subset()
+{
+  auto t0 = std::make_tuple(0, 2, 3.14);
+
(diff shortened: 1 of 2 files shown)


brycelelbach:
Submitted internal CI, but it has gaps in C++11 coverage so someone will need to test this locally. @griwes, I'm looking at you.

@griwes and @jaredhoberock to code review.

brycelelbach:
> Looks OK to me, modulo that these tests seem to use c++14 library features.

> Does the newer test build enable c++14?

Right now our main CI basically doesn't test C++11/C++14, we're working on that. Our auxiliary CI (nightly) does.

brycelelbach:
Passed internal CI, accepting.

PR #1028, opened 2019-10-26
Define THRUST_CPP_DIALECT correctly under MSVC

bjude:
MSVC doesnt define __cplusplus correctly unless a compiler flag is passed (/Zc:__cplusplus) but _MSVC_LANG is defined correctly.
To avoid users needing to pass an extra compile flag, I suggest we handle it in the cpp_dialect.h file

### DIFF
--- a/thrust/detail/config/cpp_dialect.h
+++ b/thrust/detail/config/cpp_dialect.h
@@ -16,13 +16,19 @@
 
 #pragma once
 
-#if   __cplusplus < 201103L
+#ifdef _MSC_VER
+  #define THRUST_CPP_VER _MSVC_LANG
+#else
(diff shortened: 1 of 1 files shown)


brycelelbach:
Yep, this is a clear oversight.

Running internal CI.

@griwes to code review.

brycelelbach:
Passed internal CI.

brycelelbach:
We had to revert this - what we need to do here is a bit more complicated, as MSVC isn't fully compliant (expression SFINAE) until newer versions.

Can you open an issue to discuss this further?

bjude:
Damn, i only tested using VS2019.

Issue #1030 created to track this.

PR #1042, opened 2020-01-08
Fix return type of predicate_to_integral::operator()

ahehn-nv:
Judging by the name of the functor `predicate_to_integral` and the explicit construction of `IntegralType` in the function, the return type should be `IntegralType` instead of `bool`.


jaredhoberock:
Good catch. I think this should be merged when convenient.

allisonvacanti:
Rebased on master and started internal CI under CL 28459953.

PR #1047, opened 2020-01-28
Build CUDA tests with `-rdc=true` in internal Makefiles

brycelelbach:
- Build CUDA-specific tests with `-rdc=true`; they test device-side launch,
  which needs `-rdc=true` (today, the tests fallback to invoking the serial
  algorithms on the device, which makes them slow).
- Remove old logic disabling codegen for random architectures.
- Re-enable rounding in `bench.cu`.

Bug 2808654

### DIFF
--- a/internal/benchmark/bench.cu
+++ b/internal/benchmark/bench.cu
@@ -393,7 +393,6 @@ struct experiment_driver
     );
     #endif
 
-/*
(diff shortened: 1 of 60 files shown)


brycelelbach:
Internal CI:

http://ausdvs.nvidia.com/Build_Results?virtualId=1000569766&which_page=current_build

PR #1061, opened 2020-02-25
Deprecate C++03, C++11.

allisonvacanti:
A deprecation warning is emitted when compiling thrust code against
C++03 and C++11. This warning may be disabled by defining
THRUST_IGNORE_DEPRECATED_CPP_DIALECT.

This effectively deprecates GCC < 5.x and MSVC 2015.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,7 +67,13 @@ add_definitions(-DTHRUST_DEVICE_SYSTEM=THRUST_DEVICE_SYSTEM_${THRUST_DEVICE_SYST
 set(CMAKE_CXX_STANDARD 11 CACHE STRING "The C++ version to be used.")
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+# Ignore when building thrust itself on deprecated C++ standards
+add_definitions(-DCUB_IGNORE_DEPRECATED_CPP_DIALECT)
+add_definitions(-DTHRUST_IGNORE_DEPRECATED_CPP_DIALECT)
+
 message("-- C++ Standard version: ${CMAKE_CXX_STANDARD}")
+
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -67,6 +67,9 @@ add_definitions(-DTHRUST_DEVICE_SYSTEM=THRUST_DEVICE_SYSTEM_${THRUST_DEVICE_SYST
 set(CMAKE_CXX_STANDARD 11 CACHE STRING "The C++ version to be used.")
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+# Ignore when building thrust itself on deprecated C++ standards
```
allisonvacanti: Should we leave this here, or bump the CMAKE_CXX_STANDARD above to 14 and require CI to define this as needed?

thrust/detail/config/cpp_dialect.h
```diff
@@ -14,19 +14,66 @@
...
  */
 
+/*! \file cpp_dialect.h
+ *  \brief Detect the version of the C++ standard used by the compiler.
+ */
+
 #pragma once
 
-#if   __cplusplus < 201103L
-  #define THRUST_CPP03
```
allisonvacanti: I didn't carry over the `THRUST_CPPXX` macros, since they seem to be unused by the project. Let me know if I should add them back.

jaredhoberock: The fewer macros the better, IMO.

allisonvacanti:
Thanks @jaredhoberock -- After this went up we decide to push most of these changes into CUB instead and indirectly include them here, so there will be a new version of this PR soon that shuffles things around a bit.

allisonvacanti:
Here's the new version -- it goes with thrust/cub#5.

allisonvacanti:
Merging with #1089.

Issue #1049, opened 2020-02-05
calling host function from host/device function in thrust/mr/allocator.h:109


allisonvacanti:
```
Building CUDA object CMakeFiles/thrust.example.cuda.custom_temporary_allocation.dir/examples/cuda/custom_temporary_allocation.cu.o
src/thrust/thrust/mr/allocator.h(109): warning: 
calling a __host__ function("
thrust::mr::stateless_resource_allocator<int,  ::thrust::system::cuda::detail::cuda_memory_resource<& ::cudaMalloc, & ::cudaFree,  ::thrust::cuda_cub::pointer<void> > > ::stateless_resource_allocator
") 
from a __host__ __device__ function("
thrust::mr::stateless_resource_allocator<int,  ::thrust::system::cuda::detail::cuda_memory_resource<& ::cudaMalloc, & ::cudaFree,  ::thrust::cuda_cub::pointer<void> > > ::stateless_resource_allocator [subobject]
") is not allowed
```

allisonvacanti:
This is being triggered by [`custom_temporary_allocation.cu:160`](https://github.com/thrust/thrust/blob/master/examples/cuda/custom_temporary_allocation.cu#L155-L160):

```
  thrust::host_vector<int> h_input(num_elements);
  // ...
  thrust::cuda::vector<int> d_input = h_input;
```

I've dug around for a while in the vector, allocator, and memory resource code and didn't see anything that would cause this warning.

allisonvacanti:
Closing as a duplicate of #1007.

PR #1067, opened 2020-03-06
Miscellaneous warning cleanup.

allisonvacanti:
This patch gets a clean build using recent internal nvcc and
gcc 7.4.0, with the exception of Github #1049.

# Fix CMake path for cub.

We were adding `dependencies/cub/cub` to the include line, but
including headers as `#include <cub/xxx.cuh>`. This caused builds to
pull from the system cub headers installed with the toolkit.

# Various "cannot call host from host/device" suppressions

Templated utilites `has_member_construct1` and `contiguous_storage`
were calling `__host__` only methods from `__host__ __device__` members.

# Remove spurious `__device__` markup from `device_vector`

This was also calling some tricky to find "host from h/d" warnings.

### DIFF
--- a/thrust/device_vector.h
+++ b/thrust/device_vector.h
@@ -68,14 +68,12 @@ template<typename T, typename Alloc = thrust::device_allocator<T> >
 
     /*! This constructor creates an empty \p device_vector.
      */
-    __host__
     device_vector(void)
       :Parent() {}
 
     /*! This constructor creates an empty \p device_vector.
      *  \param alloc The allocator to use by this device_vector.
      */
-    __host__
     device_vector(const Alloc &alloc)
       :Parent(alloc) {}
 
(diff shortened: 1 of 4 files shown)


thrust/device_vector.h
```diff
@@ -175,7 +175,6 @@ template<typename T, typename Alloc = thrust::device_allocator<T> >
      */
     template<typename OtherT, typename OtherAlloc>
     __host__ explicit
-    __device__
```
jaredhoberock: Do we know why this `__device__` was here?

allisonvacanti: I have no idea. The commit that added it was vague (something like "update from github"), but since `explicit` appears in between host/device, it looks like it was unintentional.

allisonvacanti:
Updated to remove `__host__` markup from `device_vector`.

allisonvacanti:
New shelve: 28115230

PR #1068, opened 2020-03-11
Add managed_pointer that is compatible with STL.

allisonvacanti:
The existing `cuda::pointer` uses a fancy reference that overloads
`operator&`, and some STL implementations misbehave when that operator
does not return the actual memory address of the object.

Since universal_memory_resource allocates memory that works on both host
and device, we need to be able to use these types with stl containers,
such as std::vector.

This patch adds a managed_pointer implementation that behaves like
`cuda::pointer`, but returns a regular c++ reference, allowing
the thrust universal allocator to work with STL containers.

### DIFF
--- a/thrust/system/cuda/memory_resource.h
+++ b/thrust/system/cuda/memory_resource.h
@@ -22,6 +22,7 @@
 
 #include <thrust/mr/memory_resource.h>
 #include <thrust/system/cuda/detail/guarded_cuda_runtime_api.h>
+#include <thrust/system/cuda/detail/managed_memory_pointer.h>
 #include <thrust/system/cuda/pointer.h>
 #include <thrust/system/detail/bad_alloc.h>
 #include <thrust/system/cuda/error.h>
@@ -86,7 +87,7 @@ namespace detail
         thrust::cuda::pointer<void> >
         device_memory_resource;
     typedef detail::cuda_memory_resource<detail::cudaMallocManaged, cudaFree,
-        thrust::cuda::pointer<void> >
+        detail::managed_memory_pointer<void> >
         managed_memory_resource;
     typedef detail::cuda_memory_resource<cudaMallocHost, cudaFreeHost,
         thrust::host_memory_resource::pointer>
--- a/testing/cuda/managed_memory_pointer.cu
+++ b/testing/cuda/managed_memory_pointer.cu
@@ -1,0 +1,141 @@
+#include <thrust/detail/config.h>
+
+#if THRUST_CPP_DIALECT >= 2011
+
+#  include <unittest/unittest.h>
+
+#  include <thrust/allocate_unique.h>
+#  include <thrust/memory/detail/device_system_resource.h>
+#  include <thrust/mr/allocator.h>
+#  include <thrust/type_traits/is_contiguous_iterator.h>
+
+#  include <numeric>
+#  include <vector>
+
+namespace
+{
+
+template <typename T>
+using allocator =
+  thrust::mr::stateless_resource_allocator<T, thrust::universal_memory_resource>;
+
+// The managed_memory_pointer class should be identified as a
+// contiguous_iterator
+THRUST_STATIC_ASSERT(
+  thrust::is_contiguous_iterator<allocator<int>::pointer>::value);
+
+template <typename T>
+struct some_object {
+  some_object(T data)
+      : m_data(data)
+  {}
+
+  void setter(T data) { m_data = data; }
+  T getter() const { return m_data; }
+
+private:
+  T m_data;
+};
+
+} // namespace
+
+template <typename T>
+void TestAllocateUnique()
+{
+  // Simple test to ensure that pointers created with universal_memory_resource
+  // can be dereferenced and used with STL code. This is necessary as some
+  // STL implementations break when using fancy references that overload
+  // `operator&`, so universal_memory_resource uses a special pointer type that
+  // returns regular C++ references that can be safely used host-side.
+
+  // These operations fail to compile with fancy references:
+  auto pRaw = thrust::allocate_unique<T>(allocator<T>{}, 42);
+  auto pObj =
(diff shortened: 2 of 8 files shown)


testing/cuda/universal_pointer.cu
```diff
@@ -0,0 +1,32 @@
...
+  // can be dereferenced and used with STL code. This is necessary as some
+  // STL implementations break when using fancy references that overload
+  // `operator&`, so universal_memory_resource uses a special pointer type that
+  // returns regular C++ references that can be safely used host-side.
+
+  // These operations fail to compile with fancy references:
+  auto p0 = thrust::allocate_unique<T>(universal_allocator<T>{});
+  p0.get();
+  *p0;
+  std::vector<T, universal_allocator<T>> v0;
```
griwes: Is this enough? Should we try instantiating some of the more common member functions that actually try to invoke operations on the value and pointer types, and not just the default constructor?

allisonvacanti:
Internal shelve: 28110928.

thrust/system/cuda/memory_resource.h
```diff
@@ -82,11 +83,123 @@ namespace detail
         return ::cudaMallocManaged(ptr, bytes, cudaMemAttachGlobal);
     }
 
+    /*! A version of thrust::cuda_cub::pointer that uses c++ references instead
+     * of thrust::cuda::reference. This is to allow managed memory pointers to
+     * be used with host-side code in standard libraries that are not compatible
+     * with proxy references.
+     */
+    template <typename T>
+    class managed_pointer
```
griwes: Why is this class here and not in pointer.h?

testing/cuda/universal_pointer.cu
allisonvacanti: These were enough to expose the compile failures with `cuda::pointer`, so it's enough in that sense. If there's more API that you think should be tested here I'm happy to add more.

thrust/system/cuda/memory_resource.h
allisonvacanti: It seemed more like an implementation detail of `managed_memory_resource`, so I left it here. It seems like it could fit either place to me.

testing/cuda/universal_pointer.cu
allisonvacanti: I think a better way to test this would be to use a `std::vector` with the universal allocator as part of the vector unit test macros, that would give much better coverage of this.

I'll get that working and update this PR when it's done.

jaredhoberock:
> Please call it universal_pointer.

I think the right name is `managed_ptr`. It's allocated by `cudaMallocManaged`. "Universal" doesn't mean anything to CUDA programmers. The `_ptr` suffix should also be used for consistency.

Others have implemented this functionality in the past, and it's been consistently referred to as "managed" pointer.

griwes:
`..._pointer` is the right naming for this, as an analogy to the `pointer` class template in the same namespace (device_ptr is a different abstraction layer).

brycelelbach:
Is this only in the cuda namespace?

Sent from Nine<http://www.9folders.com/>
________________________________

allisonvacanti:
It's in a cuda specific `detail` namespace as an implementation detail of the cuda `managed_memory_resource`.

jaredhoberock:
OK, if it's in a detail namespace, it doesn't really matter what it's called :-)

brycelelbach:
We should have a top-level way of spelling the pointer type returned by `thrust::universal_allocator`; it should be called `thrust::universal_pointer` or `thrust::universal_ptr` (@griwes will have an opinion about which of those two names is more correct, I suspect the later).

brycelelbach:
Should we have a `managed_ptr` in each backend?

allisonvacanti:
Ya'll seem to have much stronger opinions on this than I do, so let's see if we can find a color for this bikeshed that we'll all be happy with 😄If I'm following correctly:

- Jared: It doesn't matter if since it's a detail, but prefer `managed_...` to `universal_...`
- Michal: Just make sure it's `..._pointer` and not `...ptr` to prevent confusion.
- Bryce: `managed_...` makes it sound like it does resource management and might be confusing.
- Allie: 🤷‍♀ 

How about we go with `detail::managed_memory_pointer<T>`, I think that would be agreeable with everyone?

allisonvacanti:
> We should have a top-level way of spelling the pointer type returned by `thrust::universal_allocator`; it should be called `thrust::universal_pointer` or `thrust::universal_ptr` (@griwes will have an opinion about which of those two names is more correct, I suspect the later).

As it stands, `thrust::universal_allocator<T>` doesn't exist. It currently has to be spelled out explicitly, as `thrust::mr::stateless_resource_allocator<T, thrust::universal_memory_resource>`. Perhaps we should add `universal_allocator<T>` typedefs to simplify things for users?

And if we had that, I think `thrust::universal_allocator<T>::pointer` would be sufficient for spelling the type. I don't know if we really need to add a specific pointer typedef for this to `thrust::`.

jaredhoberock:
If the purpose of this allocator is to be useful in any Thrust backend, I'd prefer it simply be named `thrust::allocator` (if it must be named at all).

That naming makes it obvious what it does when CUDA is not involved.

allisonvacanti:
Only concern I have there is that we're already using the term "universal" outside of cuda, eg. `thrust::universal_memory_resource` in `thrust/memory/detail/device_system_resource.h`. In that context, `universal_allocator` is more consistent IMO.

brycelelbach:
Yes, we need a top level alias for the allocator and for the pointer type.

I'd suggest a quick meeting to discuss this.

testing/cuda/universal_pointer.cu
allisonvacanti: I've added some basic validation of using a `std::vector<T, universal_allocator>` with STL algorithms here instead.

allisonvacanti:
@brycelelbach this is blocked on your change request.

New Shelve: 28165473

PR #1070, opened 2020-03-12
Add unit test for `inclusive_scan` with User Defined Type

codereport:
This is the corresponding unit test that will test for the `struct` + `inclusive_scan` fixed here: https://github.com/thrust/cub/pull/13 For more details, read the description / fix for that bug

### DIFF
--- a/testing/scan.cu
+++ b/testing/scan.cu
@@ -648,3 +648,27 @@ void TestExclusiveScanWithBigIndexes()
 
 DECLARE_UNITTEST(TestExclusiveScanWithBigIndexes);
 
+#if THRUST_CPP_DIALECT >= 2011
+
+struct Int {
+    int i{};
+    __host__ __device__ explicit Int(int num) : i(num) {}
+    __host__ __device__ Int() : i{} {}
+    __host__ __device__ Int operator+(Int const& o) const { return Int{this->i + o.i}; }
+};
+
+void TestInclusiveScanWithUserDefinedType()
+{
+    thrust::device_vector<Int> vec(5, Int{1});
+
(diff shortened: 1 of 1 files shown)


testing/scan.cu
```diff
@@ -640,11 +640,31 @@ void TestExclusiveScanWithBigIndexesHelper(int magnitude)
...
+{
+    thrust::device_vector<Int> vec(5, Int{1});
+
+    thrust::inclusive_scan(
+        thrust::device,
+        vec.cbegin(),
+        vec.cend(),
+        vec.begin());
+
+    ASSERT_EQUAL(static_cast<Int>(vec.back()).i, 15);
```
allisonvacanti: I feel like I'm missing something here, but shouldn't the result be 5, not 15?

allisonvacanti:
These new changes will need to be wrapped in 

```
#if THRUST_CPP_DIALECT >= 2011
// ...
#endif // c++11
```

since pre-C++11 won't be able to support `{}` initialization.

testing/scan.cu
codereport: Yes! I originally had an `std::iota` in the test but decided to remove it! Forgot that change the 15 to 5. Fixed!

codereport:
> These new changes will need to be wrapped in
> 
> ```
> #if THRUST_CPP_DIALECT >= 2011
> // ...
> #endif // c++11
> ```
> 
> since pre-C++11 won't be able to support `{}` initialization.

Done :)

allisonvacanti:
Rebased and added submodule update to use thrust/cub#13.

Internal CI launched for shelve 28141734.

PR #1073, opened 2020-03-17
Optimize cuda_cub::__copy::device_to_device for contiguous trivial copies.

allisonvacanti:
This was code path was routing all copies, trivial or not, through `transform` with an identity functor.

Now, do a static check for whether the copy is trivial (iterators are contiguous and values are trivially copyable) and use `thrust::cuda_cub::trivial_copy_device_to_device` when possible.

Shelve 28136272

### DIFF
--- a/thrust/system/cuda/detail/internal/copy_device_to_device.h
+++ b/thrust/system/cuda/detail/internal/copy_device_to_device.h
@@ -29,6 +29,10 @@
 
 
 #if THRUST_DEVICE_COMPILER == THRUST_DEVICE_COMPILER_NVCC
+
+#include <thrust/advance.h>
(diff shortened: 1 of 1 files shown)


griwes:
Is this logic consistent with the logic of `thrust::async::copy`?

brycelelbach:
@allisonvacanti is this good to go?

allisonvacanti:
This is pretty old and neglected at this point. It's low priority, so I was planning to wait until CI was stable before returning to it.

allisonvacanti:
Doesn't build after rebase anymore, marking as draft until I can fix it.

allisonvacanti:
Closing this for now as it is stale and low priority. I'll come back to it at some point.

PR #1082, opened 2020-03-30
Adding CUB Version Check

Balnian:
- Added exact version match check of CUB to Thrust (if CUB version is not exactly the same as the Thrust version, a compilation error will occur)
- Added escape hatch preprocessor define to avoid the version checking

Bug 2880957

### DIFF
--- a/thrust/system/cuda/config.h
+++ b/thrust/system/cuda/config.h
@@ -71,3 +71,10 @@
 #define THRUST_CUB_NS_PREFIX namespace thrust {   namespace cuda_cub {
 #define THRUST_CUB_NS_POSTFIX }  }
 
+#ifndef THRUST_IGNORE_CUB_VERSION_CHECK
+#include <thrust/version.h>
+#include <cub/version.cuh>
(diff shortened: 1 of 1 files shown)


thrust/system/cuda/config.h
```diff
@@ -71,3 +71,10 @@
 #define THRUST_CUB_NS_PREFIX namespace thrust {   namespace cuda_cub {
 #define THRUST_CUB_NS_POSTFIX }  }
 
+#ifndef THRUST_IGNORE_CUB_VERSION_CHECK
+#include <thrust/version.h>
+#include <cub/version.cuh>
+#if THRUST_VERSION != CUB_VERSION
+#error Thrust version must match cub version exactly. Define THRUST_IGNORE_CUB_VERSION_CHECK to ignore.
```
brycelelbach, line 78: Let's change the text to: "The version of CUB in your include path is not compatible with this release of Thrust. Define THRUST_IGNORE_CUB_VERSION_CHECK to ignore this."

PR #1088, opened 2020-04-02
Allow replace algorithms with functors with non-const call operators

dkolsen-pgi:
The CUDA back end for thrust::replace_copy_if would fail to compile if
called with a predicate that had a non-const function call operator.
Fix the problem by changing thrust::cuda_cub::__replace::new_value_if_f's
function call operators to be non-const.  The operators don't need to be
const member functions because there are no const objects of type
new_value_if_f.  Having the operators be const unnecessarily requires
that the function call operator of the embedded predicate be const.

### DIFF
--- a/thrust/system/cuda/detail/replace.h
+++ b/thrust/system/cuda/detail/replace.h
@@ -64,14 +64,14 @@ namespace cuda_cub {
 
       template<class T>
       OutputType THRUST_DEVICE_FUNCTION
-      operator()(T const &x) const
+      operator()(T const &x)
       {
         return pred(x) ? new_value : x;
       }
 
       template<class T, class P>
       OutputType THRUST_DEVICE_FUNCTION
-      operator()(T const &x, P const& y) const
+      operator()(T const &x, P const& y)
       {
         return pred(y) ? new_value : x;
       }


dkolsen-pgi:
Here is a test program that fails with the current Thrust and is fixed by this change:

```
#include <thrust/replace.h>
struct always_true {
        __device__ bool operator()(int x) { return true; }
};
int main() {
        thrust::device_vector<int> input;
        thrust::device_vector<int> output;
        thrust::replace_copy_if(thrust::cuda::par,
                input.begin(), input.end(), output.begin(), always_true(), 1);
}
```
```
$ nvcc replace.cu
.../include/thrust/system/cuda/detail/replace.h(69): error: no instance of function "always_true::operator()" matches the argument list and object (the object has type qualifiers that prevent a match)
            argument types are: (const int)
            object type is: const always_true
```

allisonvacanti:
Do you have a usecase where a non-const call operator is required, or is this just for convenience? I believe that the current behavior is intentional, but need @jaredhoberock to confirm.

Since the predicate functor is copied into the device's environment but not back to the host, any changes to the functor during device execution will not be visible from the host. Requiring the call operator to be `const` is a way of enforcing this -- without this requirement users may try to write functors that cannot be supported across backends.

For a somewhat contrived example:

```
struct always_true_and_count {
  int num_executions{0};
  __device__ bool operator()(int)
  {
    ++num_executions;
    return true;
  }
}
```

`num_executions` may be non-zero after the algorithm executes on tbb, but it would always be zero on CUDA, and this sort of inconsistency is what we want to avoid. By requiring that `operator()` is `const`, these sorts of subtle bugs are avoided.

jaredhoberock:
> Do you have a usecase where a non-const call operator is required, or is this just for convenience? I believe that the current behavior is intentional, but need @jaredhoberock to confirm.

I'm not sure if the standard weighs in on this issue. Sequential algorithms allow non-const callables, but it makes sense to mutate their state in sequence.  I think Thrust should either allow non-const callables in all algorithms (unless there is some weird circumstance where it is impossible) or prohibit them everywhere to enforce a consistent rule.

I lean in favor of allowing non-const to work. Forcing programmers to decorate their functors with additional keywords that don't do any real work seems pedantic in this case. If they try to do any state mutation in their functors, it's up to them to make sure that it makes sense for their use case.

dkolsen-pgi:
This issue was uncovered by Intel's parallel algorithms test suite, not by a real world example.  The C++ standard doesn't give implementations the freedom to forbid user-supplied non-const callables.  (Thrust algorithms aren't standard conforming, but it is nice to avoid unnecessary differences.)

If I remember correctly, replace_copy_if was the only Thrust algorithm where the Intel test suite found this issue.  Other Thrust algorithms accept non-const callables without problems.

Function objects that actually change their state are problematic in any parallel algorithm, even when no GPU is involved.  There will be multiple copies of the callable (one per thread), or the callable will be shared and accessed simultaneously from multiple threads.  And callables are passed by value, even in sequential algorithms, so getting the modified state back out is difficult.  There is nothing special about thrust::replace_copy_if that makes a non-const callable more problematic there than anywhere else.

For the reasons of standards conformance and consistency, I think this is a good change.

allisonvacanti:
Ah, I was under the impression that thrust already prohibited non-const callables everywhere after a similar issue popped up on another MR. I stand corrected :)

Agreed that consistency with the other algorithms and the standard are more important here. I'll land this patch soon.

allisonvacanti:
Internal CI submitted: shelve 28210844

PR #1089, opened 2020-04-03
Deprecate C++03, C++11, MSVC < 2017, GCC < 5.0

allisonvacanti:
Deprecate everything before C++14, MSVC 2017, and GCC 5.0.

Includes many fixes to get things building on MSVC and assorted config consistency / modernization updates. See commit message for full details.


thrust/detail/config/cpp_dialect.h
```diff
@@ -14,19 +14,37 @@
...
-  #define THRUST_CPP_DIALECT 2017
-#endif
+#include <thrust/detail/config/deprecated.h>
+
+#ifndef THRUST_CPP_DIALECT
+
+// MSVC does not define __cplusplus correctly. _MSVC_LANG is used instead
+// (MSVC 2015U3+ only)
+#  ifdef _MSVC_LANG
+#    define THRUST___CPLUSPLUS _MSVC_LANG
```
griwes, line 30: Rename, this name is reserved for implementations of ISO C++ (two consecutive underscores).

thrust/detail/config/exec_check_disable.h
```diff
@@ -28,7 +28,7 @@
 #if defined(__CUDACC__) && !defined(__NVCOMPILER_CUDA__) && \
     !(defined(__CUDA__) && defined(__clang__))
 
-#define __thrust_exec_check_disable__ #pragma nv_exec_check_disable
+#define __thrust_exec_check_disable__ _Pragma("nv_exec_check_disable")
```
griwes, line 31: Yeah, I looked at this recently and wondered whether the `#pragma` form is kosher. Is this something that actually broke _somewhere_, or just a stylistic fix to make this actually behave more like a declaration annotation than a directive?

thrust/detail/config/cpp_dialect.h
allisonvacanti: Will do. We have other macros that violate this rule, but I agree we shouldn't be adding more.

thrust/detail/config/exec_check_disable.h
allisonvacanti: It didn't break anything for me *yet*, but `clang-format` likes to reformat code from

```
__thrust_exec_check_disable__
__host__ __device__
void foo()
{
  // ...
}
```

to

```
__thrust_exec_check_disable__ __host__ __device__ void foo()
{
  // ...
}
```

which would consume the `__host__ __device__ void foo()` bit as part of the `#pragma`.

allisonvacanti:
Renamed the macro and made the same change in cub as part of thrust/cub#21.

allisonvacanti:
New shelve 28219812

internal/benchmark/bench.cu
```diff
@@ -5,6 +5,8 @@
 #include <thrust/reduce.h>
 #include <thrust/scan.h>
```
brycelelbach, line 7: Let's not have a newline between `<thrust/detail/config.h>` and the other Thrust headers.

testing/mr_pool.cu
```diff
@@ -1,8 +1,11 @@
 #include <unittest/unittest.h>
+
```
brycelelbach, line 2: Let's not have a newline between `<thrust/detail/config.h>` and the other Thrust headers.

thrust/detail/complex/catrig.h
```diff
@@ -588,7 +589,7 @@ inline double real_part_reciprocal(double x, double y)
  * Re(catanh(z)) = x/|z|^2 + O(x/z^4)
  *    as z -> infinity, uniformly in x
  */
-#if __cplusplus >= 201103L || !defined _MSC_VER
+#if THRUST_CPP_DIALECT >= 2011 || !defined _MSC_VER
```
brycelelbach, line 592: Replace `!defined _MSC_VER`  with `THRUST_HOST_COMPILER != THRUST_HOST_COMPILER_MSVC`.

thrust/detail/complex/catrig.h
```diff
@@ -752,7 +753,7 @@ inline complex<double> asin(const complex<double>& z){
   return detail::complex::casin(z);
 }
   
-#if __cplusplus >= 201103L || !defined _MSC_VER
+#if THRUST_CPP_DIALECT >= 2011 || !defined _MSC_VER
```
brycelelbach, line 756: Replace `!defined _MSC_VER`  with `THRUST_HOST_COMPILER != THRUST_HOST_COMPILER_MSVC`.

thrust/detail/complex/catrig.h
```diff
@@ -773,7 +774,7 @@ inline complex<double> asinh(const complex<double>& z){
   return detail::complex::casinh(z);
 }
   
-#if __cplusplus >= 201103L || !defined _MSC_VER
+#if THRUST_CPP_DIALECT >= 2011 || !defined _MSC_VER
```
brycelelbach, line 777: Replace `!defined _MSC_VER`  with `THRUST_HOST_COMPILER != THRUST_HOST_COMPILER_MSVC`.

thrust/detail/complex/catrigf.h
```diff
@@ -386,7 +387,7 @@ inline float real_part_reciprocal(float x, float y)
   return (x / (x * x + y * y) * scale);
 }
 
-#if __cplusplus >= 201103L || !defined _MSC_VER
+#if THRUST_CPP_DIALECT >= 2011 || !defined _MSC_VER
```
brycelelbach, line 390: Replace `!defined _MSC_VER`  with `THRUST_HOST_COMPILER != THRUST_HOST_COMPILER_MSVC`.

thrust/detail/complex/catrigf.h
```diff
@@ -467,7 +468,7 @@ inline complex<float> asin(const complex<float>& z){
   return detail::complex::casinf(z);
 }
 
-#if __cplusplus >= 201103L || !defined _MSC_VER
+#if THRUST_CPP_DIALECT >= 2011 || !defined _MSC_VER
```
brycelelbach, line 471: Replace `!defined _MSC_VER`  with `THRUST_HOST_COMPILER != THRUST_HOST_COMPILER_MSVC`.

thrust/detail/complex/catrigf.h
```diff
@@ -488,7 +489,7 @@ inline complex<float> asinh(const complex<float>& z){
   return detail::complex::casinhf(z);
 }
 
-#if __cplusplus >= 201103L || !defined _MSC_VER
+#if THRUST_CPP_DIALECT >= 2011 || !defined _MSC_VER
```
brycelelbach, line 492: Replace `!defined _MSC_VER`  with `THRUST_HOST_COMPILER != THRUST_HOST_COMPILER_MSVC`.

thrust/detail/config/namespace.h
```diff
@@ -14,24 +14,11 @@
...
 
-
-#pragma once
-
-//functions to support blocking
-
+/*! \namespace thrust
+ *  \brief \p thrust is the top-level namespace which contains all Thrust
+ *         functions and types.
+ */
```
brycelelbach, line 20: I think I'd prefer to have this live in `version.h` as it does today - so maybe get rid of this new file and add this back to `version.h`.

thrust/detail/contiguous_storage.h
```diff
@@ -16,6 +16,8 @@
 
 #pragma once
 
+#include <thrust/detail/config.h>
+
```
brycelelbach, line 20: Remove the newline here.

thrust/detail/preprocessor.h
```diff
@@ -1172,7 +1174,7 @@
   #define THRUST_CURRENT_FUNCTION __FUNC__
 #elif defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 199901)
   #define THRUST_CURRENT_FUNCTION __func__
-#elif defined(__cplusplus) && (__cplusplus >= 201103)
+#elif THRUST_CPP_DIALECT >= 2011
```
brycelelbach, line 1177: This header by design includes nothing and is supposed to be self-contained (so that it can be used in other config code/stuff that is early in the dependency chain.

We don't need MSVC logic for `__cplusplus` here: MSVC has the `__FUNCSIG__` extension, which is checked for earlier in the #`ifdef` logic here.

So let's remove all the changes from this file, including the addition of the `#include`.

allisonvacanti:
New shelve 28221412

allisonvacanti:
Updated with a ton of changes that get this to build on MSVC2015.

A commit with *just* the new changes can be found here: https://github.com/allisonvacanti/thrust/commit/1c9440ea98c45b1530363cfe254a9cf8e4cd1aa8

Relaunched CI under shelve 28231909.

allisonvacanti:
Merged with the -Werror and deprecation PRs to check for incompatible CI builds. New shelve is 28232673.

brycelelbach:
It turns out that MSVC 2015 lies about being C++14 compliant. It doesn't have extended constexpr, but `_MSVC_LANG` does report C++14.

- https://gcc.godbolt.org/z/Lc23fF
- https://docs.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=vs-2019
- https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=vs-2019

So what do we want to do about this?

/polls 'Use "C++14 minus extended `constexpr`" as our base' 'Deprecate MSVC 2015 as well and fix `THRUST_CPP_DIALECT` to report MSVC as C++11 instead of C++14'

allisonvacanti:
I vote for the second option (and am currently testing MSVC 2017 and 2019).

allisonvacanti:
This last push updates the deprecation checks to give an `#error` for C++ < 14, MSVC < 2015, Clang < 6.0, GCC < 5.0. These can all be ignored by defining `THRUST_IGNORE_DEPRECATED_CPP_DIALECT` and `CUB_IGNORE_DEPRECATED_CPP_DIALECT`.

This patch also gets everything building on MSVC 2017 and 2019 using C++14 with the async algorithms enabled, and all tests pass (with the exception of #1098, that still needs to be tracked down).

New shelve: 28243163

testing/alignment.cu
```diff
@@ -210,14 +210,31 @@ void test_aligned_type()
...
     ASSERT_EQUAL(THRUST_ALIGNOF(type), Align);
     ASSERT_EQUAL(thrust::detail::alignment_of<type>::value, Align);
 }
 
+template <std::size_t Len, std::size_t Align>
+void test_aligned_storage_instantiation(thrust::detail::false_type /* Align is invalid */)
+{
+  std::cerr << "Skipping test_aligned_storage_instantiation<Len=" << Len
+            << ", Align=" << Align << "> (MaxAlign="
+            << THRUST_ALIGNOF(thrust::detail::max_align_t) << ").\n";
```
brycelelbach, line 226: Let's probably just make this a comment.

thrust/detail/complex/catrig.h
```diff
@@ -601,7 +602,12 @@ complex<double> catanh(complex<double> z)
   y = z.imag();
   ax = fabs(x);
   ay = fabs(y);
-  
+
+  // MSVC needs to pull this in from the std namespace
+#if THRUST_HOST_COMPILER == THRUST_HOST_COMPILER_MSVC
+  using std::atanh;
+#endif
```
brycelelbach, line 609: I think we probably ought to try adding a math.h include instead, or just try removing this now that the complex header properly includes math.h.

thrust/detail/execute_with_dependencies.h
```diff
@@ -189,7 +189,7 @@ struct execute_with_allocator_and_dependencies
         return std::move(dependencies);
     }
 
-    typename std::remove_reference<Allocator>::type&
+    typename std::add_lvalue_reference<Allocator>::type
     __host__
```
brycelelbach, line 193: We should probably move `__host__` before the return type.

brycelelbach: Let's open an issue and do this as a separate commit to avoid blocking this change from landing.

thrust/detail/complex/catrig.h
brycelelbach: Let's open an issue and do this as a separate commit to avoid blocking this change from landing.

testing/alignment.cu
brycelelbach: Let's open an issue and do this as a separate commit to avoid blocking this change from landing.

brycelelbach:
Make sure to put the "Reviewed-by:" line in this commit (one for each reviewer).

thrust/detail/complex/catrig.h
allisonvacanti: Ah apparently `thrust/detail/complex/c99math.h` does the equivalent of

```
namespace thrust::detail::complex { using ::atanh; }
```

for every platform but MSVC. Adding this on MSVC fixes the compile error and the complex tests are building and passing locally. I'll push the new PR in a sec.

PR #1102, opened 2020-04-15
Add rudimentary CMake support for Feta

brycelelbach:


CMakeLists.txt
```diff
@@ -499,6 +539,13 @@ foreach (THRUST_TEST_SOURCE IN LISTS THRUST_TESTS)
     thrust_testframework
     ${THRUST_ADDITIONAL_LIBRARIES})
 
+  if ("Feta" STREQUAL "${CMAKE_CUDA_COMPILER_ID}")
+    set_target_properties(${THRUST_TEST}
+      PROPERTIES CUDA_RESOLVE_DEVICE_SYMBOLS OFF)
+    set_target_properties(${THRUST_TEST_RDC}
+      PROPERTIES COMPILE_FLAGS "-ta:nordc")
```
dkolsen-pgi, line 546: I think `-ta:nordc` should be `-gpu=nordc`

CMakeLists.txt
```diff
@@ -526,8 +573,15 @@ foreach (THRUST_TEST_SOURCE IN LISTS THRUST_TESTS)
       thrust_testframework
       ${THRUST_ADDITIONAL_LIBRARIES})
 
-    set_target_properties(${THRUST_TEST_RDC}
-      PROPERTIES CUDA_SEPARABLE_COMPILATION ON)
+    if ("Feta" STREQUAL "${CMAKE_CUDA_COMPILER_ID}")
+      set_target_properties(${THRUST_TEST_RDC}
+        PROPERTIES CUDA_RESOLVE_DEVICE_SYMBOLS OFF)
+      set_target_properties(${THRUST_TEST_RDC}
+        PROPERTIES COMPILE_FLAGS "-ta:rdc")
```
dkolsen-pgi, line 580: I think `-ta:rdc` should be `-gpu=rdc`

CMakeLists.txt
```diff
@@ -624,6 +678,13 @@ foreach (THRUST_EXAMPLE_SOURCE IN LISTS THRUST_EXAMPLES)
   target_link_libraries(${THRUST_EXAMPLE}
     ${THRUST_ADDITIONAL_LIBRARIES})
 
+  if ("Feta" STREQUAL "${CMAKE_CUDA_COMPILER_ID}")
+    set_target_properties(${THRUST_EXAMPLE}
+      PROPERTIES CUDA_RESOLVE_DEVICE_SYMBOLS OFF)
+    set_target_properties(${THRUST_EXAMPLE}
+      PROPERTIES COMPILE_FLAGS "-ta:nordc")
```
dkolsen-pgi, line 685: `-ta:nordc` --> `-gpu=nordc`

CMakeLists.txt
```diff
@@ -647,8 +708,15 @@ foreach (THRUST_EXAMPLE_SOURCE IN LISTS THRUST_EXAMPLES)
     target_link_libraries(${THRUST_EXAMPLE_RDC}
       ${THRUST_ADDITIONAL_LIBRARIES})
 
-    set_target_properties(${THRUST_EXAMPLE_RDC}
-      PROPERTIES CUDA_SEPERABLE_COMPILATION ON)
+    if ("Feta" STREQUAL "${CMAKE_CUDA_COMPILER_ID}")
+      set_target_properties(${THRUST_EXAMPLE_RDC}
+        PROPERTIES CUDA_RESOLVE_DEVICE_SYMBOLS OFF)
+      set_target_properties(${THRUST_EXAMPLE_RDC}
+        PROPERTIES COMPILE_FLAGS "-ta:rdc")
```
dkolsen-pgi, line 715: `-ta:rdc` --> `-gpu=rdc`

dkolsen-pgi:
I don't know Cmake, so I can't do a meaningful review of most of the change.  But I did spot what looks like several incorrect compiler arguments.

CMakeLists.txt
brycelelbach: Fixed.

CMakeLists.txt
brycelelbach: Fixed.

CMakeLists.txt
brycelelbach: Fixed.

brycelelbach:
Can everyone please re-review, I've made some updates.

CMakeLists.txt
```diff
@@ -1,9 +1,9 @@
 cmake_minimum_required(VERSION 3.8)
 
-project(Thrust CXX)
+project(Thrust NONE)
```
griwes, line 3: This is going to be broken for non-CUDA backends, no? Because the only `enable_language(CXX)` I can see below is inside the block for the CUDA backend.

CMakeLists.txt
```diff
@@ -1,9 +1,9 @@
...
 
 set(THRUST_SOURCE ${CMAKE_SOURCE_DIR})
-include(cmake/common_variables.cmake)
+set(THRUST_FILECHECK_DATA_PATH "${THRUST_SOURCE}/internal/test")
```
griwes, line 6: This is wrong, please put it back where it was. `common_variables.cmake` is also used by thrust/cmake/run_example.cmake. There's a reason why it was like it was.

cmake/common_variables.cmake
```diff
@@ -1 +0,0 @@
-set(THRUST_FILECHECK_DATA_PATH "${THRUST_SOURCE}/internal/test")
```
griwes, line 1: Please put it back here, see the comment on line 6 of CMakeLists.txt.

brycelelbach: Done.

CMakeLists.txt
brycelelbach: What's the reason?

Done.

CMakeLists.txt
brycelelbach: Yah good point, fixing.

CMakeLists.txt
```diff
@@ -70,14 +70,46 @@ set(CMAKE_CXX_EXTENSIONS OFF)
...
+      unset(CMAKE_CUDA_HOST_COMPILER CACHE)
+      message(FATAL_ERROR "Thrust tests and examples require the C++ compiler"
+          " and the CUDA host compiler to be the same; to set this compiler, please"
+          " use the CMAKE_CXX_COMPILER variable, not the CMAKE_CUDA_HOST_COMPILER"
+          " variable.")
+    endif ()
+    set(CMAKE_CUDA_HOST_COMPILER ${CMAKE_CXX_COMPILER})
+  endif ()
+
+  enable_language(CXX)
```
griwes, line 101: This is very wrong; because we don't initialize the C++ language before forcing the host compiler, `CMAKE_CXX_COMPILER` is empty if the C++ compiler is set via an environment variable, resulting in `CMAKE_CUDA_HOST_COMPILER` ending up as the default host compiler (because it's empty) later when CUDA is enabled.

brycelelbach: Fixed, see update.

CMakeLists.txt
```diff
@@ -70,14 +70,48 @@ set(CMAKE_CXX_EXTENSIONS OFF)
...
+    if (NOT "${CMAKE_CUDA_HOST_COMPILER}" STREQUAL "")
+      unset(CMAKE_CUDA_HOST_COMPILER CACHE)
+      message(FATAL_ERROR "You are using Feta as your CUDA C++ compiler, but have"
+          " specified a different host ISO C++ compiler; Feta acts as both, so"
+          " please unset the CMAKE_CUDA_HOST_COMPILER variable.")
+    endif ()
+    set(CMAKE_CXX_COMPILER ${CMAKE_CUDA_COMPILER})
+  endif () 
+
+  enable_language(CXX)
```
allisonvacanti, line 89: Why not just have the user specify `CMAKE_CXX_COMPILER` correctly when they configure? Changing the compiler variables like this is a really odd pattern for CMake, typically these would be specified to cmake:

```
cmake -DCMAKE_CUDA_COMPILER="...." -DCMAKE_CXX_COMPILER="..." /path/to/src
```

Out of curiosity, can Feta cuda objects not be mixed with other c++ compilers?

allisonvacanti: I'd rather just see an error message if we need to enforce these to be the same. The `*_COMPILER` variables are special to CMake -- if they change, cmake will freak out, drop all custom configuration option, and reconfigure from scratch. That behavior makes me nervous about trying to be too clever here.

CMakeLists.txt
```diff
@@ -97,7 +131,14 @@ if ("CUDA" STREQUAL "${THRUST_DEVICE_SYSTEM}")
   set(THRUST_HIGHEST_COMPUTE_ARCH 75)
   set(THRUST_KNOWN_COMPUTE_ARCHS 30 32 35 50 52 53 60 61 62 70 72 75)
 
-  option(THRUST_DISABLE_ARCH_BY_DEFAULT "If ON, then all CUDA architectures are disabled on the initial CMake run." OFF)
+  if ("Feta" STREQUAL "${CMAKE_CUDA_COMPILER_ID}")
```
allisonvacanti, line 134: If Feta cannot support multiple archs at once, we can use different logic to specifying the target arch under feta:

```cmake
set(THRUST_COMPUTE_ARCH "30" CACHE STRING "Target compute architecture")
set_property(CACHE THRUST_COMPUTE_ARCH PROPERTY STRINGS ${THRUST_KNOWN_COMPUTE_ARCHS})
```

That will add a CMake option `THRUST_COMPUTE_ARCH` that will allow selection from the values in `THRUST_KNOWN_COMPUTE_ARCHS` when using ccmake/cmake-gui, while limiting the selection to a single value.

allisonvacanti:
I really dislike the pattern of changing the compiler variables programmatically for the reasons mentioned inline. But since we already do that, I wouldn't hold this PR up over it, we can address that later in the eventual CMake refactor.

The other suggestion about the arch option is a "nice to know" low priority QOL change that can be skipped for now. I assume most of these hacks are temporary until we get proper CMake support for Feta, so if it works for feta and doesn't break existing usecases, I'm ok with merging.

CMakeLists.txt
brycelelbach: Okay sounds fine.

brycelelbach:
> I really dislike the pattern of changing the compiler variables programmatically for the reasons mentioned inline. But since we already do that, I wouldn't hold this PR up over it, we can address that later in the eventual CMake refactor.

Until CMake has first class support for Feta, we probably can't avoid this.

CMakeLists.txt
brycelelbach: No, you can't mix cheeses. Feta objects are not compatible with Cheddar (aka NVCC) objects or GoUDA objects.

brycelelbach: Feta really ought to be both the CUDA compiler and the C++ compiler, at least when compiling Thrust.

CMakeLists.txt
brycelelbach: How would this work with NVCC? You use this to just target one compute arch? I'm not sure I like the idea of having different styles for different compilers.

allisonvacanti: Probably best to leave this as is then.

CMakeLists.txt
```diff
@@ -404,6 +480,11 @@ target_include_directories(
   PRIVATE ${PROJECT_SOURCE_DIR}/testing
 )
 
+if ("Feta" STREQUAL "${CMAKE_CUDA_COMPILER_ID}")
```
brycelelbach, line 483: We need to only take this branch if the device backend is CUDA.

Issue #1101, opened 2020-04-14
sort() compile failure under CPP and OMP backends


hwinkler:
Commit 1669350 introduced a change in `thrust/detail/allocator/temporary_allocator.inl` line 155 that causes code calling `sort()` and `sort_by_key()` to fail compiling under CPP and OMP backends.

See  ref https://github.com/thrust/thrust/commit/1669350bcbc026e2df10ab75bbc4f088761024d1#diff-089d87b68c34e207ae7ce4ecf499cc36R55

To repro:
```c++
#include <thrust/device_vector.h>
#include <thrust/sort.h>
void foo()
{
    thrust::device_vector<float> f (10, 1.0f);
    thrust::sort(f.begin(), f.end());
}
```

Compile like:
```
/usr/lib/cuda-10.1/bin/nvcc -DTHRUST_DEVICE_SYSTEM=THRUST_DEVICE_SYSTEM_CPP --compile  bug.cu
```
to see error messages

```
./thrust/thrust/detail/allocator/temporary_allocator.inl(55): error: name followed by "::" must be a class or namespace name
          detected during:
            instantiation of "thrust::detail::allocator_traits<Alloc>::pointer thrust::detail::allocator_traits<Alloc>::allocate(thrust::detail::allocator_traits<Alloc>::allocator_type &, thrust::detail::allocator_traits<Alloc>::size_type) [with Alloc=thrust::detail::no_throw_allocator<thrust::detail::temporary_allocator<float, thrust::system::cpp::detail::tag>>]" 
./thrust/thrust/detail/contiguous_storage.inl(206): here
            instantiation of "void thrust::detail::contiguous_storage<T, Alloc>::allocate(thrust::detail::contiguous_storage<T, Alloc>::size_type) [with T=float, Alloc=thrust::detail::no_throw_allocator<thrust::detail::temporary_allocator<float, thrust::system::cpp::detail::tag>>]" 
./thrust/thrust/detail/contiguous_storage.inl(64): here
            instantiation of "thrust::detail::contiguous_storage<T, Alloc>::contiguous_storage(thrust::detail::contiguous_storage<T, Alloc>::size_type, const thrust::detail::contiguous_storage<T, Alloc>::allocator_type &) [with T=float, Alloc=thrust::detail::no_throw_allocator<thrust::detail::temporary_allocator<float, thrust::system::cpp::detail::tag>>]" 
./thrust/thrust/detail/temporary_array.inl(76): here
[...]

PR #1103, opened 2020-04-15
Fix regression of temporary_allocator with non-CUDA back ends

dkolsen-pgi:
A recent cleanup of functions that have different code for host and device
accidentally broke temporary_allocator for non-CUDA back ends.  An #if
condition that protects a piece of code that only works with the CUDA back
end was incorrectly removed.  This fix adds that condition back in.

Fix issue #1101

### DIFF
--- a/thrust/detail/allocator/temporary_allocator.inl
+++ b/thrust/detail/allocator/temporary_allocator.inl
@@ -51,7 +51,7 @@ __host__ __device__
         throw thrust::system::detail::bad_alloc("temporary_buffer::allocate: get_temporary_buffer failed");
       #endif
     } else {
-      #if THRUST_INCLUDE_DEVICE_CODE
+      #if THRUST_INCLUDE_DEVICE_CODE && THRUST_DEVICE_SYSTEM == THRUST_DEVICE_SYSTEM_CUDA
         thrust::system::cuda::detail::terminate_with_message("temporary_buffer::allocate: get_temporary_buffer failed");
       #endif
     }


allisonvacanti:
Should `THRUST_INCLUDE_DEVICE_CODE` only be set when using the cuda device? Maybe we should fix this at the macro definition and rename the macro to `THRUST_INCLUDE_CUDA_DEVICE_CODE`. This would disambiguate between thrust's device system and cuda device code.

Thoughts?

dkolsen-pgi:
Tweaking the definition of `THRUST_INCLUDE_DEVICE_CODE` was my first thought.  But this place in `temporary_allocator` is the only place in Thrust code that has a special check for `THRUST_DEVICE_SYSTEM_CUDA`.  Everywhere else generates device code normally even in `THRUST_DEVICE_SYSTEM_CPP` mode.  So after looking into it more, I didn't want to change the general macro, and instead retained the special case code that had been within `temporary_allocator`.

allisonvacanti:
Sounds reasonable, LGTM. I'll land this sometime tomorrow 👍

griwes:
Now that I've seen this change, it seems a bit unfortunate that we are using the word "device" to mean different things in different contexts (in the context of Thrust types, we use it to mean "in the domain of the device backend", while in the context of the macros, we use it to mean "on the GPU" instead). It's probably not something we should be concerned about right now, but we should think about this.

@hwinkler can you please confirm that this patch solves the issue?

allisonvacanti:
Shelve 28262454.

hwinkler:
@griwes the patch checks out here. Compiles our codebase for CPP and OMP.

hwinkler:
wait... cancel that approval... sorry, I spoke too quickly.
please stand by... our own tests failing under cuda

hwinkler:
Sorry for the scare. All is well. The patch does check out fine. 

Testing our release build works great. The cause of the scare was, I ran the tests under our debug switches, and we've always had certain unit tests that fail under that config.

allisonvacanti:
Thanks for confirming, @hwinkler!

PR #1108, opened 2020-04-17
Get rid of a GCC 9 warning about deprecated generation of copy ctors.

griwes:
A bunch of these have been cleaned up a while ago, but since we don't
have GCC 9 CI yet, we've missed two new cases added by the recent change
to support Feta.

Internal P4 shelve CL: 28268875, CI running.


brycelelbach:
I want to use this opportunity to call out the fact that the terrible codename is catching on :)

griwes:
I'm going with it only to avoid increased confusion. :P

PR #1109, opened 2020-04-17
Opt in to deprecations in DVS

brycelelbach:
This won't make a lot of sense without the other part, but this disables adding the macros which opt out of the deprecation warnings to our command line. Note that separately we add them for Thrust/CUB for older compilers (GCC 4.8 and MSVC 2015).


brycelelbach:
@griwes has verbally consented to this.

PR #1118, opened 2020-04-22
Make submodule path global

### LINKED ISSUE #1119: Cloning thrust from a forked repository fails to update submodule.
Steps to Reproduce:

1. Fork `thrust` but not `cub` to your own user. 
2. Clone `thrust` from the forked user
3. Do a `git submodule update --init`

Expected result: `git submodule update` should work.
Result : `git submodule update` fails

jayeshbadwaik:
This commit sets `cub` repository as the global cub repository, allowing people with forks to clone and update submodules without failing. This fixes #1119

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "cub"]
 	path = dependencies/cub
-	url = ../cub.git
+	url = git@github.com:thrust/cub.git


brycelelbach:
Hi Jayesh,

Unfortunately, we can't do this just yet. The submodule structure has to be relative so that we can check out the repository internally - @griwes can explain better than I.

We're working on a plan for migrating to a saner organizational structure in the near future, which should solve this issue.

jayeshbadwaik:
No problem. :-) Feel free to close the issue if you so desire.

Issue #1115, opened 2020-04-21
`thrust::async::copy(thrust::device.after(thing), ...)` fails with a `no_state` exception because the policy is copied twice and then moved from twice


jeng1220:
The original report is #651, but the fix to the issue has been lost. The internal nvbug is 200501674. 

I also tried new Thrust API with CUDA 10.2 but it still crashed.
```c++
#include <thrust/host_vector.h>
#include <thrust/device_vector.h>
#include <thrust/async/copy.h>

int main()
{
  int N = 5000;
  thrust::host_vector<float> a(N);
  thrust::device_vector<float> b(N);
  thrust::device_event e(thrust::new_stream);
  e = thrust::async::copy(thrust::device.after(e), a.begin(), a.end(), b.begin());
  e.wait();
  return 0;
}
```
Even doing D2D memcpy encountered same error:
```c++
#include <thrust/device_vector.h>
#include <thrust/async/copy.h>

int main()
{
  int N = 5000;
  thrust::device_vector<float> a(N);
  thrust::device_vector<float> b(N);
  thrust::device_event e(thrust::new_stream);
  e = thrust::async::copy(thrust::device.after(e), a.begin(), a.end(), b.begin());
  e.wait();
  return 0;
}
```
The error message is:
```sh
terminate called after throwing an instance of 'thrust::event_error'
  what():  no_state: an operation that requires an event or future to have a stream or content has been performed on a event or future without either, e.g. a moved-from or default constructed event or future (anevent or future may have been consumed more than once)
```

jeng1220:
add @jaredhoberock to increase visibility

brycelelbach:
This happens because the one-policy overload of `async::copy` copies its single policy argument and uses it as both the "to" and "from" policy when calling the two-policy overload of `async::copy`. This is problematic because the stream-stealing code in `thrust::future` extracts the stream from the event in the first copy of the policy, leaving the second copy in a moved-from state, leading to the `no_state`  exception.

My solution is to change the one-policy overload to assume its single policy argument is the "to" policy and infer the "from" policy from iterators (which is what is done for both policies in the no-policy overload). I think this is the best thing to do - the one-policy overload of copy has always been a little vague about what it means, since copy is the one operation in Thrust that can happen between hosts and devices.

brycelelbach:
> The original report is #651, but the fix to the issue has been lost. The internal nvbug is 200501674.

[...]

PR #1131, opened 2020-05-01
Fix the one-policy overload of `thrust::async::copy` to not copy the policy

brycelelbach:
See #1115 for an explanation.


brycelelbach:
@griwes please review

brycelelbach:
Folks, please review, let's sneak this in to 1.9.10.

CL 28359146

PR #1143, opened 2020-05-12
Add a unit test for the convenient caching allocator in detail.

griwes:
Internal shelved CL: 28381691. CI running.

I was trying to figure out if this test can be cleverer in any way, but haven't figured out a good way to do that. The template is there just so that I don't have to utter the actual allocator type there; we can clean that up to just use `auto` once we actually drop C++03.

### DIFF
--- a/testing/caching_allocator.cu
+++ b/testing/caching_allocator.cu
@@ -1,0 +1,23 @@
+#include <unittest/unittest.h>
+
+#include <thrust/detail/config.h>
+#include <thrust/detail/caching_allocator.h>
+
+template<typename Allocator>
+void test_implementation(Allocator alloc)
+{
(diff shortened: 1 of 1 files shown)


testing/caching_allocator.cu
```diff
@@ -0,0 +1,23 @@
...
+    typedef typename Allocator::pointer Ptr;
+
+    Ptr p = Traits::allocate(alloc, 123);
+    Traits::deallocate(alloc, p, 123);
+
+    Ptr p2 = Traits::allocate(alloc, 123);
+    ASSERT_EQUAL(p, p2);
+}
+
+void TestCachingAllocator()
```
brycelelbach, line 19: Let's call this `TestSingleDeviceTLSCachingAllocator`?

PR #1144, opened 2020-05-12
Fix "namespace uses itself"

allisonvacanti:
iterator_category_to_traversal.h(36): warning C4515: 'detail':
  namespace uses itself

No need to bring the `thrust::detail` namespace into `thrust::detail`.


allisonvacanti:
CI: 28383933

PR #1145, opened 2020-05-12
When cleaning up type names in `unittest::base_class_name`, only call `std::string::replace` if we found the substring we are looking to replace.

brycelelbach:
We disabled demangling for Feta in 1aab840d04bf9ddbccdfab91a8089102317a18b2 -> typeid strings no longer contained "<" -> the `find_first_of` that searched for "<" in the strings started returning NPOS -> the `replace` call was passed NPOS -> a whole bunch of Feta tests failed.

### DIFF
--- a/testing/unittest/testframework.h
+++ b/testing/unittest/testframework.h
@@ -285,10 +285,13 @@ inline std::string base_class_name(const std::string& name)
   // if the name begins with "class ", chop it off
   chop_prefix(result, "class ");
 
-  // chop everything including and after first "<"
-  return result.replace(result.find_first_of("<"),
-                        result.size(),
-                        "");
+  const std::size_t first_lt = result.find_first_of("<");
+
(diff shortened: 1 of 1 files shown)


testing/unittest/testframework.h
```diff
@@ -285,10 +285,15 @@ inline std::string base_class_name(const std::string& name)
...
 
-  // chop everything including and after first "<"
-  return result.replace(result.find_first_of("<"),
-                        result.size(),
-                        "");
+  const std::size_t first_lt = result.find_first_of("<");
+
+  if (first_lt < result.size())
+      // chop everything including and after first "<"
+      return result.replace(result.find_first_of("<"),
```
griwes, line 292: Reuse `first_lt` here?

brycelelbach: Oh right.

Issue #1153, opened 2020-05-18
copy_if fails if copied objects have non-trivial constructors.


hwinkler:
`thrust::copy_if` and probably other functions fail to copy structs having a non-trivial assignment operator. Functions in `thrust/system/cuda/detail/copy_if.h` and in `cub/block/block_exchage.cuh` invoke the assignment operator to copy to temporary uninitialized memory,  but the destination item has never been constructed and the bits are random uninitialized memory. A non-trivial assignment operator expecting a valid `this` object will behave incorrectly.

I will submit a couple of PRs real soon.

```c++
#include <thrust/copy.h>
#include <thrust/device_vector.h>
#include <thrust/mr/allocator.h>
#include <iostream>

// This struct will only properly assign if its magic field is
// this certain number.
const int MAGIC = 923390;

struct X
{
    int field;
    int magic;
    __host__ __device__ X()
    {
        magic = MAGIC;
        field = 0;
    }
    __host__ __device__ X(int f)
    {
        magic = MAGIC;
        field = f;
    }

    // Omit this assignment operator to make the copy_if succeed.
    
    #if 1
    X& __host__ __device__ operator=(const X& x)
    {
        // To really copy over x's field value, require we have magic value set.
        // Since copy_if copies to uninitialized bits, the field will never be 923390.

        if (magic == MAGIC){
            field = x.field;
        } 
        return *this;
    }
    #endif
};

struct always_true {
    __host__ __device__ bool operator() (const X&) {
        return true;
    };
};


int main(void)
{
    thrust::device_vector<X> a(10, X(99));
    thrust::device_vector<X> b(10);

    thrust::copy_if(a.begin(), a.end(), b.begin(), always_true());
 
    int result = 0;
    for (int i =0; i < 10; i++){
        X ha(a[i]);
        X hb(b[i]);
        int ia = ha.field;
        int ib = hb.field;

        // The ia are all 99.
        // The ib should be too.
        if (ib != ia) {
            std::cout << "bad element in output "
              << " expected: " << ia
              << " got: " << ib
              << std::endl;
              result = 1;
        }
    }

    return result;  
}

```

allisonvacanti:
Thanks for the report and PRs! I'll take a look.

hwinkler:
[...]

PR #1154, opened 2020-05-18
Use placement new to construct item on uninitialized memory.

hwinkler:
Formerly used the assignment operator to copy to uninitialized
memory. But a non-trivial assignment operator requires the destination
object be in a valid state. So use placement new to construct the item
on the uninitialized bits.

Partial fix for #1153
Also a similar fix to come in CUB.

### DIFF
--- a/testing/copy.cu
+++ b/testing/copy.cu
@@ -1,6 +1,8 @@
 #include <unittest/unittest.h>
 #include <thrust/copy.h>
 
+#include <array>
+#include <algorithm>
 #include <list>
 #include <iterator>
 #include <thrust/sequence.h>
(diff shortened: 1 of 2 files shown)


hwinkler:
How do I build the tests?

Looks like I have to persuade it to generate some dependencies, but running `generate_mk.py` needs some arguments that I don't know how to provide.

I've poked around a little in the Makefile. `make depend` doesn't exist. Just a little file telling people how to build the tests would be helpful. Thanks. (Sorry if it's documented and I've missed it.)

```
CCBIN=/usr/bin/g++-8 PYTHON=python3  make
#### CCBIN         : /usr/bin/g++-8
#### CCBIN VERSION : g++-8 (Ubuntu 8.4.0-3ubuntu2) 8.4.0 Copyright (C) 2018 Free Software Foundation, Inc. This is free software; see the source for copying conditions.  There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
#### CXX_STD       : c++14
Makefile:192: /built/__agnostic/thrust/mk/dependencies.mk: No such file or directory
make: *** No rule to make target '/built/__agnostic/thrust/mk/dependencies.mk'.  Stop.
```

allisonvacanti:
Oh, sorry about that -- ignore any makefiles/python scripts/scons stuff you see. That's there to work with NVIDIA's amazing internal CI.

Developers should just point cmake at the root of the thrust directory.

@brycelelbach I can't find a document describing thrust's config / build process. Do we have one? If not, we should add one sooner than later.

hwinkler:
Thanks Allison. I got the tests built and I've added the regression test as you outlined.

This test will fail, though, unless you also merge the CUB PR  https://github.com/thrust/cub/pull/32, which it looks like you're on the way to doing.

allisonvacanti:
Better late than never, I've added contribution instructions in #1160, but it sounds like you got it working already!

This looks good to me. I'll take it from here and get this tested on NVIDIA's CI and eventually merge it in. Just a heads up, at some point I'll squash this down to a single commit before merging it to ease integration with our internal perforce repo.

Thanks for the patch!

allisonvacanti:
Testing this and thrust/cub#32 internally under CL 28430725.

allisonvacanti:
Rerunning CI on 28449497.

allisonvacanti:
Running more CI 28454197

allisonvacanti:
Tests are passing internally. Thanks for the patch!

PR #1157, opened 2020-05-18
Update .gitignore to exclude my build directories

brycelelbach:


allisonvacanti:
FYI, per-user stuff can just go in `thrust/.git/info/exclude` or an account-wide `git/ignore` file (see [this](https://git-scm.com/docs/gitignore)). I find it a little easier to maintain these sort of exclusions through those mechanisms since changing them doesn't require a PR.

brycelelbach:
Fair point but then I have to remember to do that on every client.

Issue #1075, opened 2020-03-22
Compilation of copy.cu fails when compiling Thrust with OpenMP


anates:
I'm trying to compile thrust with OpenMP (after not having a GPU available atm.), but the compilation fails for the file copy.cu in testing with the error
```
[ 76%] Building CXX object CMakeFiles/thrust.test.copy.dir/testing/copy.cu.o
In file included from ~/Downloads/git-files/thrust/thrust/iterator/iterator_traits.h:113,
                 from ~/Downloads/git-files/thrust/thrust/detail/type_traits/pointer_traits.h:23,
                 from ~/Downloads/git-files/thrust/thrust/type_traits/is_contiguous_iterator.h:27,
                 from ~/Downloads/git-files/thrust/thrust/type_traits/is_trivially_relocatable.h:19,
                 from ~/Downloads/git-files/thrust/thrust/detail/complex/complex.inl:20,
                 from ~/Downloads/git-files/thrust/thrust/complex.h:1031,
                 from ~/Downloads/git-files/thrust/testing/unittest/assertions.h:3,
                 from ~/Downloads/git-files/thrust/testing/unittest/unittest.h:6,
                 from ~/Downloads/git-files/thrust/testing/copy.cu:1:
~/Downloads/git-files/thrust/thrust/iterator/detail/iterator_traits.inl: In instantiation of 'struct thrust::iterator_traversal<only_set_when_expected_it>':
~/Downloads/git-files/thrust/thrust/system/omp/detail/copy.inl:104:69:   required from 'OutputIterator thrust::system::omp::detail::copy(thrust::system::omp::detail::execution_policy<System1>&, InputIterator, InputIterator, OutputIterator) [with DerivedPolicy = thrust::system::omp::detail::par_t; InputIterator = thrust::counting_iterator<long long int>; OutputIterator = only_set_when_expected_it]'
~/Downloads/git-files/thrust/thrust/detail/copy.inl:37:14:   required from 'OutputIterator thrust::copy(const thrust::detail::execution_policy_base<DerivedPolicy>&, InputIterator, InputIterator, OutputIterator) [with System = thrust::system::omp::detail::par_t; InputIterator = thrust::counting_iterator<long long int>; OutputIterator = only_set_when_expected_it]'
~/Downloads/git-files/thrust/testing/copy.cu:665:49:   required from here
[...]

PR #1158, opened 2020-05-18
Fix issues so all host/device combinations build and pass tests.

allisonvacanti:
- Add more metadata to mock specializations for testing iterator in
    testing/copy.cu.
- Add missing include to shuffle unit test.
- Specialize wrapped_function for void return types.
  - MSVC is not a fan of the pattern `return static_cast<void>(expr);`.
- Replace deprecated `tbb/tbb_thread.h` with `<thread>`.
- Fix overcounting of initial value in tbb scans.
  - Apparently reverse_join may be called before operator()
- Use `thrust::advance` instead of `+=` for generic iterators.
- Wrap the OMP flags in -Xcompiler for NVCC
- Extend ASSERT_STATIC_ASSERT skip for HOST=OMP, too
- Add missing header caught by tbb.cuda configs.
- Fix 'unsafe API' warnings in examples on MSVC: s/fopen/fstream/

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -236,6 +236,9 @@ if ("MSVC" STREQUAL "${CMAKE_CXX_COMPILER_ID}")
   # object files:
   append_option_if_available("/bigobj" THRUST_CXX_WARNINGS)
 
+  # "Oh right, this is Visual Studio."
+  add_compile_definitions("NOMINMAX")
+
   set(THRUST_TREAT_FILE_AS_CXX "/TP")
 else ()
   append_option_if_available("-Werror" THRUST_CXX_WARNINGS)
@@ -678,6 +681,13 @@ foreach (THRUST_EXAMPLE_SOURCE IN LISTS THRUST_EXAMPLES)
         PROPERTIES CUDA_SEPARABLE_COMPILATION ON)
     endif ()
   endif ()
+
+  if ("MSVC" STREQUAL "${CMAKE_CXX_COMPILER_ID}")
+    # Some examples use unsafe APIs (e.g. fopen) that MSVC will complain about
+    # unless this is set:
+    set_target_properties(${THRUST_EXAMPLE}
+      PROPERTIES COMPILE_DEFINITIONS "_CRT_SECURE_NO_WARNINGS")
+  endif()
 
   add_test(NAME ${THRUST_EXAMPLE}
     COMMAND ${CMAKE_COMMAND}
--- a/thrust/detail/function.h
+++ b/thrust/detail/function.h
@@ -24,80 +24,137 @@ namespace thrust
 namespace detail
 {
 
-
-template<typename Function, typename Result>
-  struct wrapped_function
+template <typename Function, typename Result>
+struct wrapped_function
 {
   // mutable because Function::operator() might be const
   mutable Function m_f;
 
   inline __host__ __device__
   wrapped_function()
+      : m_f()
+  {}
+
+  inline __host__ __device__
+  wrapped_function(const Function& f)
+      : m_f(f)
+  {}
+
+  __thrust_exec_check_disable__
+  template <typename Argument>
+  inline __host__ __device__
+  Result operator()(Argument& x) const
+  {
+    return static_cast<Result>(m_f(thrust::raw_reference_cast(x)));
+  }
+
+  __thrust_exec_check_disable__
+  template <typename Argument>
+  inline __host__ __device__
+  Result operator()(const Argument& x) const
+  {
+    return static_cast<Result>(m_f(thrust::raw_reference_cast(x)));
+  }
+
+  __thrust_exec_check_disable__
+  template <typename Argument1, typename Argument2>
+  inline __host__ __device__
+  Result operator()(Argument1& x, Argument2& y) const
+  {
+    return static_cast<Result>(m_f(thrust::raw_reference_cast(x),
+                                   thrust::raw_reference_cast(y)));
+  }
+
+  __thrust_exec_check_disable__
+  template <typename Argument1, typename Argument2>
+  inline __host__ __device__
+  Result operator()(const Argument1& x, Argument2& y) const
+  {
+    return static_cast<Result>(m_f(thrust::raw_reference_cast(x),
+                                   thrust::raw_reference_cast(y)));
+  }
+
+  __thrust_exec_check_disable__
+  template <typename Argument1, typename Argument2>
+  inline __host__ __device__
+  Result operator()(const Argument1& x, const Argument2& y) const
+  {
+    return static_cast<Result>(m_f(thrust::raw_reference_cast(x),
+                                   thrust::raw_reference_cast(y)));
+  }
+
+  __thrust_exec_check_disable__
+  template <typename Argument1, typename Argument2>
+  inline __host__ __device__
+  Result operator()(Argument1& x, const Argument2& y) const
+  {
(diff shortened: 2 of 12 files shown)


thrust/system/tbb/detail/scan.inl
```diff
@@ -194,32 +202,16 @@ template<typename InputIterator,
...
-    has_result_type<BinaryFunction>::value,
-    result_type<BinaryFunction>,
-    eval_if<
-      is_output_iterator<OutputIterator>::value,
-      thrust::iterator_value<InputIterator>,
-      thrust::iterator_value<OutputIterator>
-    >
+    is_output_iterator<OutputIterator>::value,
+    thrust::iterator_value<InputIterator>,
+    thrust::iterator_value<OutputIterator>
```
allisonvacanti, line 210: @jaredhoberock @brycelelbach @griwes I'm looking for your thoughts on this particular change. This was made to support this unit test:

```
    // float -> float with plus<int> operator (int accumulator)
    thrust::inclusive_scan(float_input.begin(), float_input.end(), float_output.begin(), thrust::plus<int>());
    ASSERT_EQUAL(float_output[0],  1.5);
    ASSERT_EQUAL(float_output[1],  3.0);
    ASSERT_EQUAL(float_output[2],  6.0);
    ASSERT_EQUAL(float_output[3], 10.0);
```
(https://github.com/thrust/thrust/blob/master/testing/scan.cu#L261-L266)

Note that this is doing an `inclusive_scan` with `float` input/output, but using `thrust::plus<int>` as the binary operator. The current version of the test expects the output to be at integer precision, except for the first element, which is `1.5`. This is how the CUDA backend behaves.

According to the comments here (deleted in this PR), this should be `1.0`, since the partial sums should have a valuetype matching the return type of the binary functor (in this case, `int`). This is how the TBB backend is currently behaving (until this PR).

Looking at the history of this test, it looks a lot like the expected value was changed arbitrarily instead of fixing a bug in the `cuda_cub` backend:
https://github.com/thrust/thrust/commit/3b6b107b940bc32e0c5701200f2347d6a03975ba#diff-95dc21c3de5f6e247592a56e50f1ea95

Even the comment there specifies that the accumulator should be `int`.

Currently this PR changes the TBB backend to match the expected test result of `1.5`, but now I'm wondering if we should instead be fixing all of the backends to return `1.0` for the first value of this test's output, since that appears to be the original intention of this unit test.

How do ya'll view this edge case? I see three possible takes:

a) Inclusive scan doesn't take an `initial_value`, so the first element should not have `binary_functor` applied to it. The correct value for the first output element is `1.5`.

b) The `binary_functor` returns `int`, so that type should be used to accumulate partial sums and the output should be at `int` precision. The correct value is `1.0`.

c) This edge case is silly and ill-defined, GIGO. We should remove this test case and say the result is implementation defined.

allisonvacanti:
This fixes #1075.

allisonvacanti:
CI: 28424174

thrust/system/tbb/detail/scan.inl
brycelelbach: As per [P0571](https://wg21.link/P0571), the intermediate type should be inferred from the input iterator, so it should be `float` here.

CMakeLists.txt
```diff
@@ -679,6 +682,13 @@ foreach (THRUST_EXAMPLE_SOURCE IN LISTS THRUST_EXAMPLES)
     endif ()
   endif ()
 
+  if ("MSVC" STREQUAL "${CMAKE_CXX_COMPILER_ID}")
+    # Some examples use unsafe APIs (e.g. fopen) that MSVC will complain about
+    # unless this is set:
+    set_target_properties(${THRUST_EXAMPLE}
+      PROPERTIES COMPILE_DEFINITIONS "_CRT_SECURE_NO_WARNINGS")
```
brycelelbach, line 689: Is this just for `discrete_voronoi`? Maybe we should just refactor that one, or at least open an issue tracking this.

brycelelbach: We should also document in the comment which things need this so we know the condition to remove it.

thrust/detail/function.h
```diff
@@ -24,80 +24,133 @@ namespace thrust
...
-  template<typename Argument>
-    inline __host__ __device__ Result operator()(const Argument &x) const
+  template <typename Argument>
+  inline __host__ __device__
+  Result operator()(const Argument& x) const
   {
-    // we static cast to Result to handle void Result without error
-    // in case Function's result is non-void
-    return static_cast<Result>(m_f(thrust::raw_reference_cast(x)));
+    return m_f(thrust::raw_reference_cast(x));
```
brycelelbach, line 56: For the non-void specialization, we should still `static_cast` so that we still get explicit conversions.

thrust/detail/internal_functional.h
```diff
@@ -317,11 +318,11 @@ template<typename UnaryFunction>
...
   inline __host__ __device__
-  typename enable_if_non_const_reference_or_tuple_of_iterator_references<
-    typename thrust::tuple_element<1,Tuple>::type
-  >::type
-    operator()(Tuple t)
+  void operator()(Tuple t)
   {
+    THRUST_STATIC_ASSERT_MSG(is_non_const_reference<Tuple>::value ||
+                             is_tuple_of_iterator_references<Tuple>::value,
+      "Expected a non-const reference or thrust::detail::tuple_of_iterator_references");
```
brycelelbach, line 325: As discussed on telecon, the original conditional seems to be the right one.

We should also check whether the well-formedness of this call is used in any SFINAE condition on a public interface.

thrust/system/tbb/detail/scan.inl
```diff
@@ -104,7 +104,12 @@ struct inclusive_body
 
   void reverse_join(inclusive_body& b)
   {
-    sum = binary_op(b.sum, sum);
+    // Only accumulate this functor's partial sum if this functor has been
+    // called at least once -- otherwise we'll over-count the initial value.
+    if (!first_call)
```
brycelelbach, line 109: Make sure the commit message makes a note of this fix so I'll pick it up in release notes.

allisonvacanti:
I found some issues with the `tbb.cuda` configuration while testing out multiconfig in #1159. I'll track that down and fix it in this branch before merging.

thrust/system/tbb/detail/scan.inl
allisonvacanti: Opened #1170 to track this.

allisonvacanti: This part of the patch has been taken over by #1176.

CMakeLists.txt
allisonvacanti: Replaced `fopen` with `fstream`.

thrust/detail/function.h
allisonvacanti: Restored the `static_cast`s.

thrust/detail/internal_functional.h
allisonvacanti: Reverted this change, the real fix was to add

```
namespace thrust
{
namespace detail
{
// We need this type to pass as a non-const ref for unary_transform_functor
// to compile:
template <>
struct is_non_const_reference<only_set_when_expected_it> : thrust::true_type {};
}
}
```

to testing/copy.cu.

thrust/system/tbb/detail/scan.inl
allisonvacanti: Done.

allisonvacanti:
Rerunning CI in 28484502.

PR #1161, opened 2020-05-19
Fix compilation of async algorithms when RDC is enabled.

### LINKED ISSUE #1050: Warnings and errors when enabling RDC
There are a quite a few of these when compiling with RDC enabled.

The warning at `reduce.h:275` is caused by `async_reduce_into_n` being marked up with `THRUST_RUNTIME_FUNCTION` (includes `__host__ __device__` when RDC is enabled), which calls the `__host__` only function `make_dependent_event`. The others are probably similar.

allisonvacanti:
THRUST_RUNTIME_FUNCTION includes `__device__` when RDC is enabled. Since
the async algorithms use host-only futures and events, this causes builds
to fail. Making the async entry points regular `__host__` functions fixes
this.

This fixes #1050.

### DIFF
--- a/thrust/system/cuda/detail/async/copy.h
+++ b/thrust/system/cuda/detail/async/copy.h
@@ -67,7 +67,6 @@ template <
   typename FromPolicy, typename ToPolicy
 , typename ForwardIt, typename OutputIt, typename Size
 >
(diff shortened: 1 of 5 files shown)


allisonvacanti:
CI: 28424107

PR #1156, opened 2020-05-18
Update changelog and development model document

brycelelbach:

PR #1138, opened 2020-05-11
Create CODE_OF_CONDUCT.md

codereport:
Resolves https://github.com/thrust/thrust/issues/1133

PR #1163, opened 2020-05-19
Add `CONTRIBUTING.md` and update `README.md` to include CMake build instructions.

### LINKED ISSUE #1160: Add developer build instructions to README.md
Since the Thrust repo contains files for a few different build systems, it's not clear to new developers which should be used.

We should add section to the README.md that details the build process for contributors, including dependencies, general CMake configuration steps, and Thrust-specific CMake options.

allisonvacanti:
Also added a CONTRIBUTING.md file which walks through the process of
cloning thrust, creating a fork, and pushing a pull request.

Fixes #1160.


CONTRIBUTING.md
```diff
@@ -0,0 +1,315 @@
...
+  - Controls the C++ dialect. Default: `14`
+- `THRUST_HOST_SYSTEM={CPP, TBB, OMP}`
+  - Selects the host system. Default: `CPP`
+- `THRUST_DEVICE_SYSTEM={CUDA, TBB, OMP, CPP}`
+  - Selects the device system. Default: `CUDA`
+- `THRUST_ENABLE_COMPUTE_XX={ON, OFF}`
+  - Only used when compiling with the `CUDA` device system.
+  - Controls the targeted CUDA architecture(s)
+  - Multiple options may be selected when using NVCC as the CUDA compiler.
+  - Valid values of `XX` are:
```
brycelelbach, line 105: Capitalize "valid".

brycelelbach:
We should probably merge the development model document into `CONTRIBUTING.md`. Then, after #1156 and #1138 we should have four documents in the top level of the repo:

- `README.md`
- `CONTRIBUTION.md`
- `CODE_OF_CONDUCT.md`
- `CHANGELOG.md`

README.md
```diff
@@ -71,14 +71,49 @@ int main(void)
...
+mkdir build
+cd build
+
+# Configure -- use one of the following:
+cmake ..   # Command line interface.
+ccmake ..  # ncurses GUI (Linux only)
+cmake-gui  # Graphical UI, set source/build directories in the app
+
+# Build:
+cmake --build -j .   # invokes make (or ninja, or msbuild, etc)
```
jaredhoberock, line 107: I see this error message when I run this command:

```
jhoberock@jhoberock-dt:~/dev/git/thrust-jaredhoberock/build$ cmake --build -j .
Unknown argument .
Usage: cmake --build <dir> [options] [-- [native-options]]
Options:
  <dir>          = Project binary directory to be built.
  --parallel [<jobs>], -j [<jobs>]
                 = Build in parallel using the given number of jobs. 
                   If <jobs> is omitted the native build tool's 
                   default number is used.
                   The CMAKE_BUILD_PARALLEL_LEVEL environment variable
                   specifies a default parallel level when this option
                   is not given.
  --target <tgt>..., -t <tgt>... 
                 = Build <tgt> instead of default targets.
  --config <cfg> = For multi-configuration tools, choose <cfg>.
  --clean-first  = Build target 'clean' first, then build.
                   (To clean only, use --target 'clean'.)
  --verbose, -v  = Enable verbose output - if supported - including
                   the build commands to be executed. 
  --             = Pass remaining options to the native tool.
```

jaredhoberock: Something like this worked, though:

```
cmake --build . -j 32
```

allisonvacanti: Ah, right, the `.` is an arg to `--build`, not `cmake`, and the raw `-j` is ninja specific.

I should also mention that you can just call make / ninja / etc directly, depending on the generator used. I think that's a more common pattern.

I'll update these when I get back to this.

PR #1169, opened 2020-05-27
Use placeholder expression in thrust::find

mfrancis95:


allisonvacanti:
LGTM -- builds and tests pass locally. Running CI under 28459137.

allisonvacanti:
@mfrancis95 Thanks for the patches! Could you merge #1169, #1171, and #1172 (and any similar additional updates) into one PR? It'll make it much easier on our internal testing/integration processes if these are all done in a single commit.

jaredhoberock:
I appreciate the effort in simplifying the codebase with these placeholders. The placeholder implementations are more complex than the simpler functors they replace. We should do some due diligence and check that these changes don't significantly increase compilation times. 

Also, if it turns out that these changes make the old functors obsolete, they should be eliminated with these changes.

allisonvacanti:
Good points, @jaredhoberock. I'll run my build-time benchmarks on these once they're are squashed down to a single PR/commit and report back.

PR #1176, opened 2020-05-29
Update scan deduced types to be standards-compliant.

### LINKED ISSUE #1170: Implement consistent scan accumulation per P0571
Not all backends use the same accumulator type when performing scans, and as a result, TBB scan tests are currently failing in master.

All backends and tests should be updated to determine the intermediate type as defined in [proposal P0571](https://wg21.link/P0571).

This will likely be addressed as part of #1158.

allisonvacanti:
Heads up that this patch changes the output of scan algorithms under certain edge cases. See the updated unit test for examples.

TBB's scan was implemented differently than the other backends, leading
to some failing unit tests.

This patch fixes these inconsistencies by making the following changes:

- Follow P0571's guidance regarding accumulator variable type.
  - https://wg21.link/P0571
  - The accumulator's type is now:
    - The type of the user-supplied initial value (if provided), or
    - The input iterator's value type if no initial value.
- Follow C++ standard guidance for default binary operator type.
  - https://eel.is/c++draft/exclusive.scan#1
  - Thrust binary/unary functors now specialize a default void template
    parameter. Types are deduced and forwarded transparently.
  - Updated the scan's default binary operator to the new
    `thrust::plus<>` specialization.
- The `intermediate_type_from_function_and_iterators` helper is no
  longer needed and has been removed.

Closes #1170.


thrust/functional.h
```diff
@@ -1224,10 +1393,27 @@ struct project1st
...
 
-/*! \p project2nd is a function object that takes two arguments and returns 
+template <>
+struct project1st<void, void>
+{
+  __thrust_exec_check_disable__
+  template <typename T1, typename T2>
+  __host__ __device__
+  constexpr auto operator()(T1&& t1, T2&&) const
+    noexcept(noexcept(THRUST_FWD(t1))) -> decltype(THRUST_FWD(t1))
```
dkolsen-pgi, line 1410: `noexcept(THRUST_FWD(t1))` is always true, and `decltype(THRUST_FWD(t1))` is `decltype(t1)` which is `T1&&`.  But there is also a lot of good in consistency with other similar classes even if that results in extra typing.  So I am not requesting any changes.

allisonvacanti:
Internal CL: 28484173

thrust/functional.h
```diff
@@ -139,6 +139,39 @@ struct binary_function
...
+  struct func<void>                                                            \
+  {                                                                            \
+    __thrust_exec_check_disable__                                              \
+    template <typename T>                                                      \
+    __host__ __device__                                                        \
+    constexpr auto operator()(T&& x) const                                     \
+      noexcept(noexcept(impl)) -> decltype(impl)                               \
+    {                                                                          \
+      return impl;                                                             \
+    }                                                                          \
```
brycelelbach, line 153: Need to add `is_transparent` here, which is a type alias for a valid unspecified type (http://eel.is/c++draft/arithmetic.operations.plus).

thrust/functional.h
```diff
@@ -139,6 +139,39 @@ struct binary_function
...
+  struct func<void>                                                            \
+  {                                                                            \
+    __thrust_exec_check_disable__                                              \
+    template <typename T1, typename T2>                                        \
+    __host__ __device__                                                        \
+    constexpr auto operator()(T1&& t1, T2&& t2) const                          \
+      noexcept(noexcept(impl)) -> decltype(impl)                               \
+    {                                                                          \
+      return impl;                                                             \
+    }                                                                          \
```
brycelelbach, line 167: Need to add `is_transparent` here, which is a type alias for a valid unspecified type (http://eel.is/c++draft/arithmetic.operations.plus).

thrust/system/tbb/detail/scan.inl
```diff
@@ -208,18 +208,10 @@ template<typename InputIterator,
...
-    >
-  >::type ValueType;
-  
-  typedef typename thrust::iterator_difference<InputIterator>::type Size; 
-  
+  // Use the input iterator's value type per https://wg21.link/P0571
+  using ValueType = typename thrust::iterator_value<InputIterator>::type;
+
+  using Size = typename thrust::iterator_difference<InputIterator>::type;
   Size n = thrust::distance(first, last);
```
brycelelbach, line 215: We could use `auto` here and then `decltype(n)` later when we use the size type instead.

thrust/system/tbb/detail/scan.inl
```diff
@@ -260,18 +252,10 @@ template<typename InputIterator,
...
-    >
-  >::type ValueType;
-
-  typedef typename thrust::iterator_difference<InputIterator>::type Size; 
-  
+  // Use the initial value type per https://wg21.link/P0571
+  using ValueType = InitialValueType;
+
+  using Size = typename thrust::iterator_difference<InputIterator>::type;
   Size n = thrust::distance(first, last);
```
brycelelbach, line 259: We could use `auto` here and then `decltype(n)` later when we use the size type instead.

brycelelbach:
Make sure to note in the commit message that this is a breaking change so that I will include it in the breaking changes section of the changelog for the next release.

PR #1181, opened 2020-06-07
Fix compilation  for clang cuda compiler

atchouprakov:
Few minor changes in test code to make tests work with clang as a device compiler.


atchouprakov:
@Artem-B

allisonvacanti:
This looks good to me. Running internal CI on CL 28516194.

PR #1182, opened 2020-06-08
serial_set_intersection range check

atchouprakov:
Following tests were failing when using clang cuda compiler:
* set_intersection_by_key_descending
* set_intersection_by_key
* set_intersection_descending
* set_intersection

### DIFF
--- a/thrust/system/cuda/detail/set_operations.h
+++ b/thrust/system/cuda/detail/set_operations.h
@@ -923,10 +923,11 @@ namespace __set_operations {
 #pragma unroll
       for (int i = 0; i < ITEMS_PER_THREAD; ++i)
       {
+        if(bBegin==bEnd || aBegin==aEnd) break;
         bool pA = compare_op(aKey, bKey);
         bool pB = compare_op(bKey, aKey);
 
-        // The outputs must come from A by definition of set interection.
+        // The outputs must come from A by definition of set intersection.
         output[i]  = aKey;
         indices[i] = aBegin;
 


atchouprakov:
@Artem-B

allisonvacanti:
I haven't looked too closely at this change, but found an issue while doing a screening test of this PR. The patch looks straightforward enough, but for some reason massively impacts build times.

Time needed for a clean build of `thrust.test.set_intersection`:

| Branch | MSVC 2019 + NVCC 11.0 | GCC 7.5.0 + NVCC 11.0 |
|--|--|--|
|  #1182 | 8:33 | 17:29 |
| master | 0:38 | 00:52 |

The other TUs that use set operations are similarly affected, and this causes the full build time for a 2 SM CUDA configuration on GCC 7.5.0 to go from 17 minutes on master to 35 minutes with this patch. Everything did build and test successfully, but it took about twice as long.

I haven't tracked down where / why this is happening yet, but we should dig in and address this before merging this one.

Artem-B:
My guess is that the `if` may have been interfering with the loop unrolling before. The patch may have made unrolling easier. It's possible that now the loop gets unrolled much more than it was before and that may have resulted in higher compilation time.

I would try limiting the #pragma unroll to specific unroll count and see if it helps.

atchouprakov:
It also might be that now it has to generate both branches and both include two calls to "compare_op", it should be OK to move the if below calls to compare_op. I'll try and report result.

atchouprakov:
I think @Artem-B is right, if I change it to
```
 if ((aBegin < aEnd) && (bBegin < bEnd))
        {
          break;
        }
```
the compilation is fast, but clang cuda will complain that in couldn't unroll the loop, nvcc doesn't complaint, but I suppose it cannot unroll it too.

allisonvacanti:
Jared pointed out that `nvcc --time` can help point at where things are going off in the `nvcc` build process.

But from your investigations it does sound like it's related to the unrolling. Maybe we should just get rid of the `#pragma unroll` here and see how it impacts performance. It shouldn't be a problem if nvcc is already not unrolling this loop. @jaredhoberock @brycelelbach thoughts?

I'm still laughing about such an innocent looking change resulting in a 20x build time increase. Madness.

Do ya'll notice a similar change in build time on clang cuda?

atchouprakov:
The compile time on clang doesn't vary much. On nvcc the slowest portion was `src/thrust/testing/set_intersection.cu , cicc , /tmp/tmpxft_00027ea2_00000000-6_set_intersection.cpp1.ii  , /tmp/tmpxft_00027ea2_00000000-5_set_intersection.ptx , compute_61 , nvcc , 511441.7812 , ms` 
I updated PR now it compiles fast on nvcc and also passes test on clang.

atchouprakov:
Unfortunately,  the `break` version fails the tests on clang for sm_70, but passing on sm_60... With the `if` version tests were passing on both. I'll close this PR for now, and reopen when we figure out what is going on.

Issue #1178, opened 2020-05-30
device_reference not stripped before passing values to placeholder actors in some cases


smitsyn:
I've faced a bug that I believe is caused by `thrust::device_reference<T>` in some cases not being stripped to `T` before being passed to function objects that are composed using placeholders (actors?). The minimal example I could come up with:

```
#include <thrust/device_vector.h>
#include <thrust/inner_product.h>
#include <thrust/functional.h>

void foo()
{
    using namespace thrust::placeholders;

    thrust::device_vector<float> v1, v2;

    thrust::inner_product(v1.begin(), v1.end(), v2.begin()
        , 0.0f, thrust::plus<float>{}, _1 + 1.0f);
}

```

Here, the problem is related to the last argument (BinaryOp2). The same can be demonstrated with this function object:

```
struct erroneous_thing
{
    template <typename T>
    __host__ __device__
        T operator()(const T& lhs, const T& rhs) const {
        return lhs + rhs;
    }
};
//...........
    thrust::inner_product(v1.begin(), v1.end(), v2.begin()
        , 0.0f, thrust::plus<float>{}, erroneous_thing{});
```

Somehow `erroneous_thing` is instantiated with `T` being `thrust::device_reference<float>`, while it should be float, resulting in a compile error. I haven't faced the problem with any other algorithm I've used; `thrust::transform` properly strips reference, as well as reduction part of inner_product (e.g. passing `erroneous_thing{}` as a first parameter is fine).

CUDA toolkit 10.2.

allisonvacanti:
Compiler Explorer link: https://www.godbolt.org/z/-8gpEp

allisonvacanti:
Fixed in #1237.

PR #1186, opened 2020-06-09
Use placeholder expressions

mfrancis95:
Use placeholder expression in thrust::count

Use placeholder expression in thrust::find

Use placeholder expression in thrust::mismatch

Use placeholder expression in thrust::sequence

Remove sequence_functor

Use placeholder expression in thrust::remove

Use placeholder expression in thrust::find

Use placeholder expression in thrust::replace

Use placeholder expression in thrust::replace

Use placeholder expression in thrust::replace_copy

### DIFF
--- a/thrust/system/cuda/detail/find.h
+++ b/thrust/system/cuda/detail/find.h
@@ -205,10 +205,12 @@ find(execution_policy<Derived> &policy,
      InputIt                    last,
      T const& value)
 {
+  using thrust::placeholders::_1;
+
   return cuda_cub::find_if(policy,
                         first,
                         last,
-                        thrust::detail::equal_to_value<T>(value));
+                        _1 == value);
 }
 
 
--- a/thrust/system/cuda/detail/remove.h
+++ b/thrust/system/cuda/detail/remove.h
@@ -74,8 +74,9 @@ remove(execution_policy<Derived> &policy,
        InputIt                    last,
        const T &                  value)
 {
-  thrust::detail::equal_to_value<T> pred(value);
-  return cuda_cub::remove_if(policy, first, last, pred);
+  using thrust::placeholders::_1;
+
+  return cuda_cub::remove_if(policy, first, last, _1 == value);
 }
 
 // copy
--- a/thrust/system/cuda/detail/replace.h
+++ b/thrust/system/cuda/detail/replace.h
@@ -90,12 +90,14 @@ replace(execution_policy<Derived> &policy,
         T const &                  old_value,
         T const &                  new_value)
 {
+  using thrust::placeholders::_1;
+
   cuda_cub::transform_if(policy,
                       first,
                       last,
                       first,
                       __replace::constant_f<T>(new_value),
-                      thrust::detail::equal_to_value<T>(old_value));
+                      _1 == old_value);
 }
 
 template <class Derived,
--- a/thrust/system/detail/generic/find.inl
+++ b/thrust/system/detail/generic/find.inl
@@ -45,8 +45,9 @@ InputIterator find(thrust::execution_policy<DerivedPolicy> &exec,
                    InputIterator last,
                    const T& value)
 {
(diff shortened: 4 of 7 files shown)


allisonvacanti:
Thanks @mfrancis95.

Following up the conversation in #1169, the default `cpp.cuda.cpp14` config builds in 17:05 on my machine, which is not noticeably different than before.

I found a couple missing headers while building some of the non-CUDA configurations, these should be fixed in the commit I just pushed.

This LGTM, I'll start internal CI soon.

mfrancis95:
Thanks for handling the missing headers. My goal was to try to remove the need for equal_to_value entirely, but there were a couple of cases where replacing it with placeholders wouldn't compile.

allisonvacanti:
There have been some issues reported when using placeholders (e.g. #1178) that we can better address now that C++03 has been dropped. I suspect it was a similar issue?

In any case, it was good to find out that the placeholder expression templates don't add much compilation overhead. Maybe someday we can switch over completely.

Started internal CI in CL 28521254.

mfrancis95:
Here's an example of a change that fails:

```
diff --git a/thrust/system/cuda/detail/remove.h b/thrust/system/cuda/detail/remove.h
index c590a1ad..c8560148 100644
--- a/thrust/system/cuda/detail/remove.h
+++ b/thrust/system/cuda/detail/remove.h
@@ -35,7 +35,7 @@ namespace thrust
 namespace cuda_cub {
 
 // in-place

 template <class Derived,
           class InputIt,
           class StencilIt,
@@ -124,8 +124,8 @@ remove_copy(execution_policy<Derived> &policy,
             OutputIt                   result,
             const T &                  value)
 {
-  thrust::detail::equal_to_value<T> pred(value);
-  return cuda_cub::remove_copy_if(policy, first, last, result, pred);
+  using thrust::placeholders::_1;
+  return cuda_cub::remove_copy_if(policy, first, last, result, _1 == value);
 }
 
 }    // namespace cuda_cub
```

jaredhoberock:
@allisonvacanti Thanks for verifying that the build time didn't change.

My only remaining concern is that the ad hoc functors that these placeholder expressions replace should be eliminated, if possible.

It looks like these changes eliminate `sequence_functor`, is it possible to eliminate `equal_to_value` as well?

Also, shouldn't these changes make some of the `#include`s used by these implementations superfluous? Where is `equal_to_value` defined?

mfrancis95:
@jaredhoberock see the snippet above. I think it's possible to eliminate that functor but it'll take some time to understand why it can't be substituted in a few places and how to fix it. Maybe it is related to #1178?

jaredhoberock:
@mfrancis95 Thanks. Let's not worry about `equal_to_value` in this issue, then.

I'd still like to eliminate `#include <thrust/detail/internal_functional.inl>` in the files, where possible. This is the header which defines these ad hoc functors.

For example, I think eliminating this one should be possible: https://github.com/thrust/thrust/pull/1186/files?file-filters%5B%5D=.inl#diff-45f0ebb2fcb6910d7ffc103f18bac4cbL21

mfrancis95:
Hi @jaredhoberock. After going through the files, `thrust/system/detail/generic/replace.inl` was the only file where I was able to remove that include. Other files that I modified still had a couple of instances where they were using other internal functors such as `thrust::detail::not1`.

The commit is going to be pushed up momentarily.

mfrancis95:
@allisonvacanti there's been a new commit pushed up to this branch. I would assume I did all that can be done for this particular wave of changes for now.

I'd want to see #1178 fixed before coming back and doing more.

allisonvacanti:
That sounds good to me. I'll restart the testing and land this if all goes well.

allisonvacanti:
Rebased and squashed in last push.

Running CI under 28707351

allisonvacanti:
CI looks good, I'll merge this soon.

PR #1190, opened 2020-06-09
Make out_of_memory_recovery test trigger OOM faster.

### LINKED ISSUE #1183: test.out_of_memory_recovery is extremely slow
This test takes about a minute and a half to run on a system with 32GB RAM + 32GB swap, and it renders the system unresponsive for most of that time as it slowly allocates the entire 64 GB of available memory, thrashing all other processes. We should look into ways to speed this up and play nice with other processes.

allisonvacanti:
Fixes #1183. This test is taking up the majority of the test
runtime on CPU backends, slowing eating away at RAM/swap for
two minutes while the rest of the system gets evicted from RAM
and stops responding.

Replaced the allocation loop with a single large allocation,
now the test runs in ~1ms and doesn't actually allocate
significant resources.

### DIFF
--- a/testing/out_of_memory_recovery.cu
+++ b/testing/out_of_memory_recovery.cu
@@ -1,4 +1,14 @@
 // Regression test for NVBug 2720132.
+//
+// Summary of 2720132:
+//
+// 1. The large allocation fails due to running out of memory.
+// 2. A `thrust::system::system_error` exception is thrown.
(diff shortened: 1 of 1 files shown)


allisonvacanti:
CL 28522192.

allisonvacanti:
A full multiconfig test cycle is now 45 minutes shorter 🎉 🎉 🎉

Issue #1195, opened 2020-06-12
`thrust::device.on(stream).after(event)`


brycelelbach:
`thrust::device.on(stream).after(event)` doesn't work; sort relies on doing `.after` internally, thus `thrust::async::sort(thrust::device.on(stream), ...)` doesn't work.

This is because `execute_on_stream` doesn't combine with other execution policy properties. Unfortunately making it work that way would require an exponential explosion of intermediate policies:

* `execute_with_allocator_and_stream`
* `execute_with_dependencies_and_stream`
* `execute_with_allocator_and_dependencies_and_stream`

The easiest solution is probably to just fold the functional of `execute_on_stream` into `execute_with_dependencies`, which can already express and hold streams.

The only downside with that approach is that `.on().on().on()` would leave you with a large tuple holding some useless streams.

Also there is a testing gap here.

brycelelbach:
This will fix #1232 by making the test not fail; before this is approved the PR should make sure to re-enable the test.

tomilov:
@allisonvacanti @brycelelbach Can expressions `thrust::cuda::par` and `thrust::cuda::par.on(cudaStream_t)` (and more general: `thrust::device` and `thrust::device.on`, if there is any for other backends) be of the same type? Is it possible to dispatch the difference in runtime? Currently I can't turn my library into the header only using class templates parameterized by `typename Policy`, because in the case its compilation time jumps to 6-7 minutes (for DEVICE=CUDA) from 24 seconds.

PR #1198, opened 2020-06-12
Make `execute_on_stream` inherit from `dependencies_aware_execution_policy` 

brycelelbach:
This is a short-term fix, long-term some major restructuring is needed.

### DIFF
--- a/testing/async_copy.cu
+++ b/testing/async_copy.cu
@@ -343,6 +343,8 @@ struct test_async_copy_after
       thrust::device.after(e0), d0.begin(), d0.end(), d1.begin()
     );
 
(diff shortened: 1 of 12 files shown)


allisonvacanti:
LGTM -- Have you run this through DVS?

allisonvacanti:
Actually, we should probably add at least one test for this use case. I'm assuming this is for issue #1195?

brycelelbach:
Yah agreed we need a test for this.

brycelelbach:
So this doesn't actually work as intended. The longer term solution may be needed instead.

brycelelbach:
Superseded by #1230

Issue #1132, opened 2020-05-01
Add test coverage for per-thread default streams


brycelelbach:
PR #1128 added support for using the per-thread default stream in Thrust if it has been requested via `CUDA_API_PER_THREAD_DEFAULT_STREAM`. We don't have any tests for this codepath, or in fact any multi-threaded host tests. We should probably add some.

rongou:
Great idea! I'm seeing some synchronization issues in spark with per-thread default stream enabled, but not sure if the problem is with thrust or cudf (or both). Having some test coverage would definitely help.

rongou:
@brycelelbach I think I might have some time to work on this. Do you have any idea on how to test this effectively?

brycelelbach:
Spawn a thread with `std::thread`, get the stream in the parent thread and the child thread by calling `thrust::cuda_cub::stream`, then compare them for equality?

We should probably also make this a CMake option so we can run tests with and without this.

rongou:
Ah yes I think we can easily test that we are using the correct streams. I'm also wondering if we can test the multi-threaded behavior, checking for race conditions, etc.

allisonvacanti:
As discussed in today's meeting, we want to update both the per-thread stream and RDC options as follows:

- Add a small number of dedicated tests that always enable / always disable these features (small, quick sanity checks)
- Add a global option that enables / disables these features in all targets, when applicable

This will give each build some coverage, while letting us add wider coverage with CI.

PR #1201, opened 2020-06-19
Add tests for legacy and per-thread default streams

rongou:
Some simple tests to verify we are getting the correct default streams. Also added the capability to pass in additional compile flags for tests.

Part of #1132.

@brycelelbach

### DIFF
--- a/testing/CMakeLists.txt
+++ b/testing/CMakeLists.txt
@@ -111,6 +111,14 @@ function(thrust_add_test target_name_var test_name test_src thrust_target)
   if (("OMP" IN_LIST config_systems) OR ("TBB" IN_LIST config_systems))
     set_tests_properties(${test_target} PROPERTIES RUN_SERIAL ON)
   endif()
+
+  # Check for per-test script. Script will be included in the current scope
+  # to allow custom property modifications.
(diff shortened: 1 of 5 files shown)


allisonvacanti:
The tests LGTM, but the CMake stuff is going to conflict with #1184.

We'll also need to update the Makefiles to make sure that this works with our current internal CI (which does not use CMake).

I'll take care of these issues once #1184 is merged and start testing integrating then.

Thanks for the patch!

rongou:
@allisonvacanti merged with the new cmake files. Please take another look. Thanks!

allisonvacanti:
@rongou Thanks for the updates and sorry about the long wait.

In general this looks ok, but I'll have to rethink this a bit from the CMake side since we need to support CUDA compilers besides nvcc. I'm planning to take a look at this sometime this week.

allisonvacanti:
I'm fairly sure this will work on all of the NVIDIA compilers, just waiting on the nvc++ folks to confirm what their default behavior is.

allisonvacanti:
Rebased on current `main`, restored original author metadata, and submitted DVS CL 28862904.

PR #1207, opened 2020-06-29
Legacy Makefiles: Don't force C++14 with GCC 4.8

brycelelbach:
Bug 200618218

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,8 +14,8 @@
 
 # Makefile for building Thrust unit test driver
 
-# Force C++14 mode. NVCC will ignore it if the host compiler doesn't support it.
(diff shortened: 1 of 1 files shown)


brycelelbach:
This fixes NVBug 200618218.

Makefile
```diff
@@ -15,7 +15,7 @@
 # Makefile for building Thrust unit test driver
 
 # Force C++14 mode. NVCC will ignore it if the host compiler doesn't support it.
```
allisonvacanti, line 17: This comment is out of date

brycelelbach: Fixed.

PR #1208, opened 2020-06-29
Update the new multiconfig CMake setup to support Feta

brycelelbach:
* Correctly set C++ langauge dialects for Feta.
* Pass warning flags to Feta.


CMakeLists.txt
```diff
@@ -72,6 +72,44 @@ if (NOT "Feta" STREQUAL "${CMAKE_CUDA_COMPILER_ID}")
...
+  set(CMAKE_CUDA_COMPILE_FEATURES
+    ${CMAKE_CUDA03_COMPILE_FEATURES}
+    ${CMAKE_CUDA11_COMPILE_FEATURES}
+    ${CMAKE_CUDA14_COMPILE_FEATURES}
+    ${CMAKE_CUDA17_COMPILE_FEATURES}
+    ${CMAKE_CUDA20_COMPILE_FEATURES}
+  )
+endif ()
+
+message(STATUS "CUDA compile features: ${CMAKE_CUDA_COMPILE_FEATURES}")
```
allisonvacanti, line 111: `CMAKE_CUDA_COMPILE_FEATURES` seems to list available features, not just the ones that we're currently using. It'd be confusing to print out features that aren't actually enabled in the current build.

It's also not documented until cmake 3.17, so I'm not sure this will work on all supported cmake versions.

Can you remove this line, and add

```
cmake_minimum_required(VERSION 3.17)
```

inside one of the feta checks?

brycelelbach: Done.

PR #1219, opened 2020-07-02
Create wrapper to use when including <algorithm> and <memory>

### LINKED ISSUE #1218: Create include wrappers for algorithm and memory
When compiling with NVC++, it is possible that `#include <thrust/host_vector.h>` will result in circular includes, in `thrust::host_vector` being used before it is defined, and in a cascading series of meaningless compilation errors.  The fix for this requires cooperation between NVC++ and Thrust.  On the Thrust, side all includes of `<algorithm>`, `<numeric>`, and `<memory>` need to be replaced by `<thrust/detail/algorithm.h>` (or similar) which wraps the include of the real file with a define/undef of a macro than NVC++ can look for.

dkolsen-pgi:
Accommodate compilers that use Thrust in their implementations of
C++ Standard algorithms.  To avoid cycles of include files, the
compiler needs to know when an include of an algorithms-related header
(`<algorithm>`, `<numeric>`, or `<memory>`) is coming from Thrust vs from
user code.  Create wrapper files to use when including `<algorithm>` or
`<memory>` that defines a macro that the compiler can check to know
where the include is coming from.  Change all includes of those files
within Thrust code to include the wrapper file instead.  (The same
change would also be made for `<numeric>` if Thrust files included
`<numeric>` anywhere.)

Fix #1218

### DIFF
--- a/thrust/addressof.h
+++ b/thrust/addressof.h
@@ -8,7 +8,7 @@
 #include <thrust/detail/config.h>
 
 #if THRUST_CPP_DIALECT >= 2011
-#  include <memory>
+#  include <thrust/detail/memory_wrapper.h>
 #endif
 
 namespace thrust
--- a/thrust/detail/algorithm_wrapper.h
+++ b/thrust/detail/algorithm_wrapper.h
@@ -1,0 +1,27 @@
+/*
+ *  Copyright 2020 NVIDIA Corporation
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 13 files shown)


allisonvacanti:
Was it not possible to break the include cycle?

Testing locally on nvcc now.

allisonvacanti:
CL 28707242

dkolsen-pgi:
I could not find any other way to break the include cycle.  If you are curious, I can explain the situation in more detail (though not in this P/R).

allisonvacanti:
I'll take your word for it, it can be a real pain to untangle these sorts of issues. I'm ok with just triaging this for now.

allisonvacanti:
That said -- If there's anything to know other than "these headers are too tightly coupled", can you create a new issue with those details at some point so we can revisit this later?

allisonvacanti:
CI looks good. I'll merge this soon.

PR #1222, opened 2020-07-06
CMake updates, add CUB (tests, etc) as optional build feature.

### LINKED ISSUE #1192: Optionally run CUB tests from Thrust CMake builds.
This should be limited to only when a) CUDA is requested and b) `THRUST_ENABLE_CUB_TESTING` is set.

allisonvacanti:
- Add option THRUST_INCLUDE_CUB_CMAKE that enables CUB build targets
  and test entries from Thrust builds. Default is off.
- Remove out of date compiler checks, since a more up-to-date check is
  implemented at build time.
- Rename ThrustCUDAConfig.cmake -> ThrustCudaConfig.cmake for consistency.
- Remove CheckCXX*.cmake files since these are included in cmake 3.15.
- Sanitize variable names in AppendOptionIfAvailable.
  - GCC wasn't detecting -Werror, this fixes it.
- Add `thrust.all` metatarget that just builds thrust when CUB is enabled.
- Add utility CMake script that sorts and prints per-target build times
  from `.ninja_log`.

Fixes #1192.


allisonvacanti:
See also thrust/cub#41.

CMakeLists.txt
```diff
@@ -82,24 +84,13 @@ message(STATUS "TBB system found?  ${THRUST_TBB_FOUND}")
...
-    message(FATAL_ERROR "This version of MSVC no longer supported.")
-  endif ()
-endif ()
-
-if ("GNU" STREQUAL "${CMAKE_CXX_COMPILER_ID}")
-  if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 4.4)
-    message(FATAL_ERROR "This version of GCC no longer supported.")
-  endif ()
-endif ()
-
```
brycelelbach, line 99: This moved somewhere else?

allisonvacanti: It's out of date and we check for this at compile time now.

PR #1226, opened 2020-07-10
Add a GitHub Action that mirrors `main` to `master`.

brycelelbach:


.github/workflows/mirror-main-to-master.yml
```diff
@@ -0,0 +1,17 @@
+on:
+  push:
+    branches:
+      - 'main'
+
+jobs:
+  mirror_job:
+    runs-on: ubuntu-latest
+    name: Mirror SOURCE to DEST
```
griwes, line 9: Replace the placeholders with the actual branch names.

brycelelbach: Fixed, but I didn't catch this on CUB before merging, I'll open another PR.

PR #1228, opened 2020-07-10
Include <thrust/detail/memory_wrapper.h>, not <memory>

dkolsen-pgi:
Yesterday "#include <memory>" was added to
thrust/detail/allocator/allocator_traits.inl.  That caused a regression
when compiling with NVC++, breaking #1218.  This fix changes that include
to be "#include <thrust/detail/memory_wrapper.h>".  (And it makes the
same change to a different include of <memory> that I seem to have missed
in my earlier commit.)

### DIFF
--- a/thrust/allocate_unique.h
+++ b/thrust/allocate_unique.h
@@ -16,7 +16,7 @@
 #include <thrust/detail/allocator/allocator_traits.h>
 
 #include <utility>
-#include <memory>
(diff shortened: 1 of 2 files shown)


allisonvacanti:
Ack, I should have caught that since we just merged the other fixes. I'll see if I can get a copy of nvc++ and add it to my pre-merge testing.

I'll get this landed and backported to 1.9.10-1.

PR #1238, opened 2020-07-17
Include <thrust/detail/memory_wrapper.h>, not <memory>

### LINKED ISSUE #1236: Divergent use of `<thrust/detail/memory_wrapper.h>` in some headers between Git and Perforce
These headers do not use `<thrust/detail/memory_wrapper.h>`. Should they?

There is divergence between Perforce and Git here. In Perforce, they use them.

Please resolve ASAP. If the resolution is that they shouldn't be used, then the change will need to be made manually in Perforce.

https://github.com/thrust/thrust/blob/main/thrust/allocate_unique.h#L19

https://github.com/thrust/thrust/blob/main/thrust/detail/allocator/allocator_traits.inl#L26

It would also be good to figure out how we ended up here

dkolsen-pgi:
Redo an earlier fix, which was somehow lost.  Fix #1236

### DIFF
--- a/thrust/allocate_unique.h
+++ b/thrust/allocate_unique.h
@@ -16,7 +16,7 @@
 #include <thrust/detail/allocator/allocator_traits.h>
 
 #include <utility>
-#include <memory>
+#include <thrust/detail/memory_wrapper.h>
 
 namespace thrust
 {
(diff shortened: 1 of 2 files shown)


allisonvacanti:
Backported to staging branch, was already in perforce gpgpu/rel

PR #1240, opened 2020-07-17
Revert cmake bump from stage

allisonvacanti:
@brycelelbach This is needed for PGI folks, they can't upgrade to 3.18 yet. Staging branch only.


brycelelbach:
I don't understand why there are five commits here?

brycelelbach:
See #1242.

PR #1246, opened 2020-07-24
Explicitly disable per-target CUDA_ARCHITECTURES props.

allisonvacanti:
Setting CMP0104 to OLD wasn't enough to suppress the errors about
this property empty being uninitialized. Setting the prop to OFF
explicitly silences the errors.


allisonvacanti:
Requires thrust/cub#50.

Issue #976, opened 2019-04-18
Support adding Thrust to CMake projects with `add_subdirectory`


Char-Aznable:
I have been using the github thrust for a while before the recent merge with cuda one. I manage dependency using cmake fetchcontent. After merging with the recent update in this repo, which includes a CMakeLists.txt, I found that it breaks command like add_subdirectory() because of the following error:
```
 MSVC_COMPILER_FLAGS:                                                                                                                                           
 | WARN_ALL : '/Wall'                                                                                                                                          
 | WARNINGS_AS_ERRORS : '/Wx'                                                                                                                                  
 | RELEASE : '/Ox'                                                                                                                                             
 | DEBUG : '/Zi;-D_DEBUG;/MTd'                                                                                                                                 
 | EXCEPTION_HANDLING : '/EHsc'                                                                                                                                
 | CPP : ''                                                                                                                                                    
 | OMP : '/openmp'                                                                                                                                             
 | TBB : ''                                                                                                                                                    
 | CUDA : ''                                                                                                                                                   
 | CUDA_BULK : ''                                                                                                                                              
 | WORKAROUNDS : '/DNOMINMAX;/wd4503'                                                                                                                          
 | C++03 : ''
 | C++11 : '-std=c++11'
-- Looking for C++ include pthread.h
-- Looking for C++ include pthread.h - found
-- Looking for pthread_create
-- Looking for pthread_create - found
-- Found Threads: TRUE
-- Found CUDA: /net/software/modules-sw/cuda/10.1/Linux/RHEL6/x86_64 (found version "10.1")                                                           
-- Found OpenMP_CXX: -fopenmp (found version "4.5")
-- Found OpenMP: TRUE (found version "4.5")
-- Found 49 examples
-- Found 5 examples/cuda
-- Found 4 examples/cpp_integration
-- Found 152 tests in testing
-- Found 59 tests in backend
CMake Error at build_cuda/_deps/thrust-src/testing/CMakeLists.txt:48 (add_custom_target):                                                                     
  add_custom_target cannot create target "check" because another target with
  the same name already exists.  The existing target is a custom target
  created in source directory
  "/home/aznb/mycodes/SCgenome_scbmc/build_cuda/_deps/kokkos-src".                                                                         
  See documentation for policy CMP0002 for more details.
```
Aside from having name conflicts via add_custom_target, I was expecting using thrust as a header-only library and I don't expect cmake to config building any of the test targets unless I want it so.

[...]

PR #1248, opened 2020-07-24
Allow thrust to be added to CMake projects via add_subdirectory.

allisonvacanti:
See issue #976.

Added example in `examples/cmake/add_subdir/CMakeLists.txt` that is
used for documentation and regression testing.


examples/cmake/add_subdir/dummy.cpp
```diff
@@ -0,0 +1,32 @@
+#include <thrust/detail/config.h>
```
brycelelbach, line 1: Why is this a .cpp file?

allisonvacanti: It's not part of thrust, it's part of a example project that demonstrates how to use the add_subdirectory(...) stuff to build an application that conditionally builds applications with different backends, one cpp and one cuda. I figured it be helpful to show the common trick of `#include <some_file.cpp>` from `some_file.cu` when doing that sort of thing.

allisonvacanti:
DVS CL: 28913883

PR #1251, opened 2020-08-06
Add async scan algorithms

allisonvacanti:
Prerequisite: NVIDIA/cub#210

### DIFF
--- a/testing/async/exclusive_scan/large_indices.cu
+++ b/testing/async/exclusive_scan/large_indices.cu
@@ -1,0 +1,244 @@
+#include <thrust/detail/config.h>
+
+#if THRUST_CPP_DIALECT >= 2014
+
+#include <async/test_policy_overloads.h>
+
+#include <async/exclusive_scan/mixin.h>
+
+#include <thrust/device_free.h>
+#include <thrust/device_malloc.h>
+#include <thrust/device_ptr.h>
+#include <thrust/optional.h>
+
+#include <thrust/iterator/detail/device_system_tag.h>
+#include <thrust/iterator/detail/iterator_facade_category.h>
+
+#include <cstdint>
+
+// This test is an adaptation of TestInclusiveScanWithBigIndices from scan.cu.
+
+namespace
+{
+
+// Fake iterator that asserts
+// (a) it is written with a sequence and
+// (b) a defined maximum value is written at some point
+//
+// This allows us to test very large problem sizes without actually allocating
+// large amounts of memory that would exceed most devices' capacity.
+struct assert_sequence_iterator
+{
+  using value_type      = std::int64_t;
+  using difference_type = std::int64_t;
+
+  // Defined for thrust::iterator_traits:
+  using pointer           = value_type*;
+  using reference         = assert_sequence_iterator; // weird but convenient
+  using iterator_category =
+    typename thrust::detail::iterator_facade_category<
+      thrust::device_system_tag,
+      thrust::random_access_traversal_tag,
+      value_type,
+      reference>::type;
+
+  std::int64_t expected{0};
+  std::int64_t max{0};
+  mutable thrust::device_ptr<bool> found_max{nullptr};
+  mutable thrust::device_ptr<bool> unexpected_value{nullptr};
+
+  // Should be called on the first iterator generated. This needs to be
+  // done explicitly from the host.
+  void initialize_shared_state()
+  {
+    found_max        = thrust::device_malloc<bool>(1);
+    unexpected_value = thrust::device_malloc<bool>(1);
+    *found_max        = false;
+    *unexpected_value = false;
+  }
+
+  // Should be called only once on the initialized iterator. This needs to be
+  // done explicitly from the host.
+  void free_shared_state() const
+  {
+    thrust::device_free(found_max);
+    thrust::device_free(unexpected_value);
+    found_max        = nullptr;
+    unexpected_value = nullptr;
+  }
+
+  __host__ __device__ assert_sequence_iterator operator+(difference_type i) const
+  {
+    return clone(expected + i);
+  }
+
+  __host__ __device__ reference operator[](difference_type i) const
+  {
+    return clone(expected + i);
+  }
+
+  // Some weirdness, this iterator acts like its own reference
+  __device__ assert_sequence_iterator operator=(value_type val)
+  {
+    if (val != expected)
+    {
+      printf("Error: expected %lld, got %lld\n", expected, val);
+      *unexpected_value = true;
+    }
+    else if (val == max)
+    {
+      *found_max = true;
+    }
+
+    return *this;
+  }
+
+private:
+  __host__ __device__
+  assert_sequence_iterator clone(value_type new_expected) const
+  {
+    return {new_expected, max, found_max, unexpected_value};
+  }
+};
+
+// output mixin that generates assert_sequence_iterators.
+// Must be paired with validate_assert_sequence_iterators mixin to free
+// shared state.
+struct assert_sequence_output
+{
+  struct output_type
+  {
+    using iterator = assert_sequence_iterator;
+
+    iterator iter;
+
+    explicit output_type(iterator&& it)
+        : iter{std::move(it)}
+    {
+      iter.initialize_shared_state();
+    }
+
+    ~output_type()
+    {
+      iter.free_shared_state();
+    }
+
+    iterator begin() { return iter; }
+  };
+
+  template <typename InputType>
+  static output_type generate_output(std::size_t num_values, InputType&)
+  {
+    using value_type = typename assert_sequence_iterator::value_type;
+    assert_sequence_iterator it{0,
+                                // minus one bc exclusive scan:
+                                static_cast<value_type>(num_values - 1),
+                                nullptr,
+                                nullptr};
+    return output_type{std::move(it)};
+  }
+};
+
+struct validate_assert_sequence_iterators
+{
+  using output_t = assert_sequence_output::output_type;
+
+  template <typename EventType>
+  static void compare_outputs(EventType& e,
+                              output_t const&,
+                              output_t const& test)
+  {
+    testing::async::mixin::compare_outputs::detail::basic_event_validation(e);
+
+    ASSERT_EQUAL(*test.iter.unexpected_value, false);
+    ASSERT_EQUAL(*test.iter.found_max, true);
+  }
+};
+
+//------------------------------------------------------------------------------
+// Overloads without custom binary operators use thrust::plus<>, so use
+// constant input iterator to generate the output sequence:
+struct default_bin_op_overloads
+{
+  using postfix_args_type = std::tuple< // List any extra arg overloads:
+    std::tuple<>,                       // - no extra args
+    std::tuple<uint64_t>                // - initial_value
+    >;
+
+  static postfix_args_type generate_postfix_args()
+  {
+    return postfix_args_type{std::tuple<>{}, std::tuple<uint64_t>{0}};
+  }
+};
+
+struct default_bin_op_invoker
+    : testing::async::mixin::input::constant_iterator_1<std::int64_t>
+    , assert_sequence_output
+    , default_bin_op_overloads
+    , testing::async::mixin::invoke_reference::noop
+    , testing::async::exclusive_scan::mixin::invoke_async::simple
+    , validate_assert_sequence_iterators
+{
+  static std::string description()
+  {
+    return "test large array indices with default binary operator";
+  }
+};
+
+} // anon namespace
+
+void test_large_indices_default_scan_op()
+{
+  // Test problem sizes around signed/unsigned int max:
+  testing::async::test_policy_overloads<default_bin_op_invoker>::run(1ll << 30);
+  testing::async::test_policy_overloads<default_bin_op_invoker>::run(1ll << 31);
+  testing::async::test_policy_overloads<default_bin_op_invoker>::run(1ll << 32);
+  testing::async::test_policy_overloads<default_bin_op_invoker>::run(1ll << 33);
+}
+DECLARE_UNITTEST(test_large_indices_default_scan_op);
+
+namespace
+{
+
+//------------------------------------------------------------------------------
+// Generate the output sequence using counting iterators and thrust::max<> for
+// custom operator overloads.
+struct custom_bin_op_overloads
+{
+  using postfix_args_type = std::tuple<     // List any extra arg overloads:
+    std::tuple<uint64_t, thrust::maximum<>> // - initial_value, binop
+  >;
+
+  static postfix_args_type generate_postfix_args()
+  {
+    return postfix_args_type{std::make_tuple(0, thrust::maximum<>{})};
+  }
+};
+
+struct custom_bin_op_invoker
+  : testing::async::mixin::input::counting_iterator_from_1<std::int64_t>
+    , assert_sequence_output
+    , custom_bin_op_overloads
+    , testing::async::mixin::invoke_reference::noop
+    , testing::async::exclusive_scan::mixin::invoke_async::simple
+    , validate_assert_sequence_iterators
+{
+  static std::string description()
+  {
+    return "test large array indices with custom binary operator";
+  }
+};
+
+} // namespace
+
+void test_large_indices_custom_scan_op()
+{
+  // Test problem sizes around signed/unsigned int max:
+  testing::async::test_policy_overloads<custom_bin_op_invoker>::run(1ll << 30);
+  testing::async::test_policy_overloads<custom_bin_op_invoker>::run(1ll << 31);
+  testing::async::test_policy_overloads<custom_bin_op_invoker>::run(1ll << 32);
+  testing::async::test_policy_overloads<custom_bin_op_invoker>::run(1ll << 33);
+}
+DECLARE_UNITTEST(test_large_indices_custom_scan_op);
+
+#endif // C++14
--- a/testing/async/exclusive_scan/large_types.cu
+++ b/testing/async/exclusive_scan/large_types.cu
@@ -1,0 +1,58 @@
+#include <thrust/detail/config.h>
+
+#if THRUST_CPP_DIALECT >= 2014
+
+#include <async/test_policy_overloads.h>
+
+#include <async/exclusive_scan/mixin.h>
+
+#include <unittest/special_types.h>
+
+// This test is an adaptation of TestScanWithLargeTypes from scan.cu.
+
+// Need special initialization for the FixedVector type:
+template <typename value_type>
+struct device_vector_fill
+{
+  using input_type = thrust::device_vector<value_type>;
+
+  static input_type generate_input(std::size_t num_values)
+  {
+    input_type input(num_values);
+    thrust::fill(input.begin(), input.end(), value_type{2});
+    return input;
+  }
+};
+
+template <typename value_type, typename alternate_binary_op = thrust::maximum<>>
+struct invoker
+    : device_vector_fill<value_type>
+    , testing::async::mixin::output::device_vector<value_type>
+    , testing::async::exclusive_scan::mixin::postfix_args::
+        all_overloads<value_type, alternate_binary_op>
+    , testing::async::exclusive_scan::mixin::invoke_reference::host_synchronous<
+        value_type>
+    , testing::async::exclusive_scan::mixin::invoke_async::simple
+    , testing::async::mixin::compare_outputs::assert_almost_equal_if_fp_quiet
+{
+  static std::string description()
+  {
+    return "scan with large value types.";
+  }
+};
+
+struct test_large_types
+{
+  void operator()(std::size_t num_values) const
+  {
+    using testing::async::test_policy_overloads;
+
+    test_policy_overloads<invoker<FixedVector<int, 1>>>::run(num_values);
+    test_policy_overloads<invoker<FixedVector<int, 8>>>::run(num_values);
+    test_policy_overloads<invoker<FixedVector<int, 32>>>::run(num_values);
+    test_policy_overloads<invoker<FixedVector<int, 64>>>::run(num_values);
+  }
+};
+DECLARE_UNITTEST(test_large_types);
+
+#endif // C++14
--- a/testing/async/exclusive_scan/mixin.h
+++ b/testing/async/exclusive_scan/mixin.h
@@ -1,0 +1,119 @@
+#pragma once
+
+#include <thrust/detail/config.h>
+
+#if THRUST_CPP_DIALECT >= 2014
+
+#include <thrust/async/scan.h>
+
+#include <thrust/scan.h>
+
+#include <async/mixin.h>
+
+namespace testing
+{
+namespace async
+{
+namespace exclusive_scan
+{
+
+namespace mixin
+{
+
+//------------------------------------------------------------------------------
+namespace postfix_args
+{
+
+template <typename value_type, typename alternate_binary_op = thrust::maximum<>>
+struct all_overloads
+{
+  using postfix_args_type = std::tuple<         // List any extra arg overloads:
+    std::tuple<>,                               // - no extra args
+    std::tuple<value_type>,                     // - initial_value
+    std::tuple<value_type, alternate_binary_op> // - initial_value, binary_op
+    >;
+
+  static postfix_args_type generate_postfix_args()
+  {
+    return postfix_args_type{std::tuple<>{},
+                             std::make_tuple(value_type{42}),
+                             std::make_tuple(value_type{42},
+                                             alternate_binary_op{})};
+  }
+};
+
+} // namespace postfix_args
+
+//------------------------------------------------------------------------------
+namespace invoke_reference
+{
+
+template <typename input_value_type,
+          typename output_value_type = input_value_type>
+struct host_synchronous
+{
+  template <typename InputType,
+            typename OutputType,
+            typename PostfixArgTuple,
+            std::size_t... PostfixArgIndices>
+  static void invoke_reference(InputType const& input,
+                               OutputType& output,
+                               PostfixArgTuple&& postfix_tuple,
+                               std::index_sequence<PostfixArgIndices...>)
+  {
+    // Create host versions of the input/output:
+    thrust::host_vector<input_value_type> host_input(input.cbegin(),
+                                                     input.cend());
+    thrust::host_vector<output_value_type> host_output(host_input.size());
+
+    // Run host synchronous algorithm to generate reference.
+    thrust::exclusive_scan(host_input.cbegin(),
+                           host_input.cend(),
+                           host_output.begin(),
+                           std::get<PostfixArgIndices>(
+                             THRUST_FWD(postfix_tuple))...);
+
+    // Copy back to device.
+    output = host_output;
+  }
+};
+
+} // namespace invoke_reference
+
+//------------------------------------------------------------------------------
+namespace invoke_async
+{
+
+struct simple
+{
+  template <typename PrefixArgTuple,
+            std::size_t... PrefixArgIndices,
+            typename InputType,
+            typename OutputType,
+            typename PostfixArgTuple,
+            std::size_t... PostfixArgIndices>
+  static auto invoke_async(PrefixArgTuple&& prefix_tuple,
+                           std::index_sequence<PrefixArgIndices...>,
+                           InputType const& input,
+                           OutputType& output,
+                           PostfixArgTuple&& postfix_tuple,
+                           std::index_sequence<PostfixArgIndices...>)
+  {
+    auto e = thrust::async::exclusive_scan(
+      std::get<PrefixArgIndices>(THRUST_FWD(prefix_tuple))...,
+      input.cbegin(),
+      input.cend(),
+      output.begin(),
+      std::get<PostfixArgIndices>(THRUST_FWD(postfix_tuple))...);
+    return e;
+  }
+};
+
+} // namespace invoke_async
+
+} // namespace mixin
+} // namespace exclusive_scan
+} // namespace async
+} // namespace testing
+
+#endif // C++14
--- a/testing/async/inclusive_scan/large_types.cu
+++ b/testing/async/inclusive_scan/large_types.cu
@@ -1,0 +1,58 @@
+#include <thrust/detail/config.h>
+
+#if THRUST_CPP_DIALECT >= 2014
+
+#include <async/test_policy_overloads.h>
+
+#include <async/inclusive_scan/mixin.h>
+
+#include <unittest/special_types.h>
+
+// This test is an adaptation of TestScanWithLargeTypes from scan.cu.
+
+// Need special initialization for the FixedVector type:
+template <typename value_type>
+struct device_vector_fill
+{
+  using input_type = thrust::device_vector<value_type>;
+
+  static input_type generate_input(std::size_t num_values)
+  {
+    input_type input(num_values);
+    thrust::fill(input.begin(), input.end(), value_type{2});
+    return input;
+  }
+};
+
+template <typename value_type, typename alternate_binary_op = thrust::maximum<>>
+struct invoker
+    : device_vector_fill<value_type>
+    , testing::async::mixin::output::device_vector<value_type>
+    , testing::async::inclusive_scan::mixin::postfix_args::
+        all_overloads<alternate_binary_op>
+    , testing::async::inclusive_scan::mixin::invoke_reference::host_synchronous<
+        value_type>
+    , testing::async::inclusive_scan::mixin::invoke_async::simple
+    , testing::async::mixin::compare_outputs::assert_almost_equal_if_fp_quiet
+{
+  static std::string description()
+  {
+    return "scan with large value types.";
+  }
+};
+
+struct test_large_types
+{
+  void operator()(std::size_t num_values) const
+  {
+    using testing::async::test_policy_overloads;
+
+    test_policy_overloads<invoker<FixedVector<int, 1>>>::run(num_values);
+    test_policy_overloads<invoker<FixedVector<int, 8>>>::run(num_values);
+    test_policy_overloads<invoker<FixedVector<int, 32>>>::run(num_values);
+    test_policy_overloads<invoker<FixedVector<int, 64>>>::run(num_values);
+  }
+};
+DECLARE_UNITTEST(test_large_types);
+
+#endif // C++14
--- a/testing/async/inclusive_scan/mixed_types.cu
+++ b/testing/async/inclusive_scan/mixed_types.cu
@@ -1,0 +1,109 @@
+#include <thrust/detail/config.h>
+
+#if THRUST_CPP_DIALECT >= 2014
+
+#include <async/test_policy_overloads.h>
+
+#include <async/inclusive_scan/mixin.h>
+
+// Test using mixed int/float types for:
+// - input_value_type       | (int, float)
+// - output_value_type      | (int, float)
+// - thrust::plus<T> T-type | (int, float, void>)
+//
+// The thrust::plus<T> types are covered by the
+// scan_mixed_types_overloads component.
+//
+// The testing/scan.cu TestMixedTypes test spells out the expected behavior,
+// which is defined by https://wg21.link/P0571.
+
+namespace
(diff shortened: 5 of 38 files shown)


testing/async_exclusive_scan.cu
```diff
@@ -0,0 +1,130 @@
...
+                           host_output.begin(),
+                           std::get<PostfixArgIndices>(
+                             THRUST_FWD(postfix_tuple))...);
+
+    // Copy back to device.
+    output = host_output;
+  }
+
+  // Invoke the async algorithm for a single overload as described by
+  // the prefix and postfix tuples. These tuple contains instances of any
```
allisonvacanti, line 79: s/These tuple/These tuples/

allisonvacanti:
The "large indices" tests are currently failing because temporary memory allocation fails for the larger problem sizes. This doesn't happen for the synchronous scan (which uses a custom implementation instead of just calling `cub::DeviceScan`). I'll pull those tests out before merging if I can't find a solution/workaround.

thrust/system/cuda/detail/async/exclusive_scan.h
```diff
@@ -0,0 +1,180 @@
...
+  {
+    ev = make_dependent_event(
+      std::tuple_cat(
+        std::make_tuple(std::move(content)),
+        extract_dependencies(std::move(thrust::detail::derived_cast(policy)))));
+  }
+
+  // Run scan.
+  thrust::cuda_cub::throw_on_error(
+    cub::DeviceScan::ExclusiveScan(
```
trivialfis, line 127: Sorry for the many questions.  From this https://github.com/NVIDIA/thrust/issues/1299#issuecomment-704024397, should we explicitly specify the size type via cub dispatching?

allisonvacanti: Yep, that seems to be the issue with the failing test. I'll need to update this to use those APIs instead.

allisonvacanti:
This is ready for review, but will need NVIDIA/cub#213 merged first so we can test without hitting timeouts.

testing/async/exclusive_scan/basic.cu
```diff
@@ -0,0 +1,72 @@
...
+    , testing::async::mixin::compare_outputs::assert_maybe_fuzzy_equal_quiet
+{
+  static std::string description()
+  {
+    return "basic invocation with device vectors";
+  }
+};
+
+template <typename T>
+struct TestBasic
```
brycelelbach, line 30: We really ought to unify naming conventions to `snake_case`; I prefer that new stuff at least use `snake_case`

brycelelbach: I'd prefer to have `exclusive_scan` in the name to be consistent with other tests.

brycelelbach: And perhaps call this "simple", because I think that's the convention of the other tests (instead of "basic").

testing/async/exclusive_scan/large_indices.cu
```diff
@@ -0,0 +1,239 @@
...
+#include <cstdint>
+
+// This test is an adaptation of TestInclusiveScanWithBigIndices from scan.cu.
+
+namespace
+{
+
+// Fake iterator that asserts
+// (a) it is written with a sequence and
+// (b) a defined maximum value is written at some point
```
brycelelbach, line 26: Can we leave a note here explaining this is necessary to test that larger than 32bit indices work without actually needing storage for that many indices (which would require a lot of GPU memory).

testing/async/exclusive_scan/large_indices.cu
```diff
@@ -0,0 +1,239 @@
...
+      value_type,
+      reference>::type;
+
+  std::int64_t expected{0};
+  std::int64_t max{0};
+  mutable thrust::device_ptr<bool> found_max{nullptr};
+  mutable thrust::device_ptr<bool> unexpected_value{nullptr};
+
+  // Should be called on the first iterator generated.
+  void initialize_shared_state()
```
brycelelbach, line 48: Why is not a constructor? Leave a note explaining.

testing/async/exclusive_scan/large_indices.cu
```diff
@@ -0,0 +1,239 @@
...
+  {
+    found_max        = thrust::device_malloc<bool>(1);
+    unexpected_value = thrust::device_malloc<bool>(1);
+    *found_max        = false;
+    *unexpected_value = false;
+  }
+
+  // Should be called only once on the initialized iterator.
+  void free_shared_state() const
```
brycelelbach, line 57: Why is not a destructor? Leave a note explaining.

brycelelbach: @griwes

testing/async/exclusive_scan/large_indices.cu
brycelelbach: @griwes

testing/async/exclusive_scan/large_indices.cu
```diff
@@ -0,0 +1,239 @@
...
+
+// This test is an adaptation of TestInclusiveScanWithBigIndices from scan.cu.
+
+namespace
+{
+
+// Fake iterator that asserts
+// (a) it is written with a sequence and
+// (b) a defined maximum value is written at some point
+struct assert_sequence_iterator
```
brycelelbach, line 27: I'd prefer to have a single implementation of this for both algorithms if possible to reduce code duplication.

testing/async/exclusive_scan/large_types.cu
```diff
@@ -0,0 +1,63 @@
...
+
+struct TestLargeTypes
+{
+  void operator()(std::size_t num_values) const
+  {
+    using testing::async::test_policy_overloads;
+
+    test_policy_overloads<invoker<FixedVector<int, 1>>>::run(num_values);
+
+#ifndef __QNX__ // These are excluded on QNX for scan.cu's version of this test.
```
brycelelbach, line 52: I'm not sure why this was failing only on QNX, but I think we should remove this ifdef and see what happens!

testing/async/exclusive_scan/mixin.h
```diff
@@ -0,0 +1,116 @@
...
+
+    // Copy back to device.
+    output = host_output;
+  }
+};
+
+} // namespace invoke_reference
+
+//------------------------------------------------------------------------------
+namespace invoke_async
```
brycelelbach, line 81: Can you give me an overview of the namespace/naming scheme used here?

testing/async/mixin.h
```diff
@@ -0,0 +1,645 @@
...
+//    // Generate an instance of the output:
+//    // Might be more complicated, eg. fancy iterators, etc
+//    static output_type generate_output(std::size_t num_values)
+//    {
+//      return output_type(num_values);
+//    }
+//
+//    //-------------------------------------------------------------------------
+//    //
+//    testing::async::mixin::exclusive_scan::mixin::postfix_args::all_overloads
```
brycelelbach, line 128: Can the second `mixin` be dropped here?

testing/async/test_policy_overloads.h
```diff
@@ -0,0 +1,421 @@
...
+                                    ValidateEvent &&validate,
+                                    std::size_t const num_values)
+  {
+    try
+    {
+      basic_policy_test<postfix_idx>(policy_tuple,
+                                     validate,
+                                     num_values);
+    }
+    catch (unittest::UnitTestException &exc)
```
brycelelbach, line 163: We have a bit of a bug when catching exceptions from non-CUDA backends, notably TBB and OpenMP; they wrap exceptions that occur in their threads in a special exception type (`tbb::captured_exception`, etc) which means these catch blocks won't match.

testing/async/test_policy_overloads.h
```diff
@@ -0,0 +1,421 @@
...
+                                      output_d,
+                                      postfix_tuple,
+                                      postfix_index_seq{});
+
+    // Let reference calc overlap with async testing:
+    algo_def::invoke_reference(input_ref,
+                               output_ref,
+                               postfix_tuple,
+                               postfix_index_seq{});
+
```
brycelelbach, line 257: We should probably leave a comment here explaining that waiting happens in `compare_outputs`

thrust/system/cuda/detail/async/scan.h
```diff
@@ -0,0 +1,33 @@
+/******************************************************************************
```
brycelelbach, line 1: Do we really need this header? Can we get rid of it and just use granular includes?

testing/async/mixin.h
```diff
@@ -0,0 +1,645 @@
...
+//            typename initial_value_type  = input_value_type,
+//            typename alternate_binary_op = thrust::maximum<>>
+//   struct basic_invoker
+//      : testing::async::mixin::input::device_vector<input_value_type>
+//      , testing::async::mixin::output::device_vector<output_value_type>
+//      , testing::async::exclusive_scan::mixin::postfix_args::
+//          all_overloads<initial_value_type, alternate_binary_op>
+//      , testing::async::exclusive_scan::mixin::invoke_reference::
+//          host_synchronous<input_value_type, output_value_type>
+//      , testing::async::exclusive_scan::mixin::invoke_async::basic
```
brycelelbach, line 73: We should call `basic` simple because `simple` is the name used in existing.

testing/async/mixin.h
```diff
@@ -0,0 +1,645 @@
...
+//  public:
+//
+//    static std::string description()
+//    {
+//      return "basic invocation with device vectors";
+//    }
+//
+//    //-------------------------------------------------------------------------
+//    // testing::async::mixin::input::device_vector
+//    using input_type  = thrust::device_vector<input_value_type>;
```
brycelelbach, line 105: There's an extra space here: `input_type  =`.

testing/async/mixin.h
```diff
@@ -0,0 +1,645 @@
...
+//    //-------------------------------------------------------------------------
+//    //
+//    testing::async::mixin::exclusive_scan::mixin::postfix_args::all_overloads
+//    //
+//    using postfix_args_type = std::tuple<                 // List any extra
+//    arg overloads:
+//      std::tuple<>,                                       // - no extra args
+//      std::tuple<initial_value_type>,                     // - initial_value
+//      std::tuple<initial_value_type, alternate_binary_op> // - initial_value,
+//      binary_op
```
brycelelbach, line 135: `binary_op` appears to have been line wrapped, it should be on the previous line.

testing/async/mixin.h
```diff
@@ -0,0 +1,645 @@
...
+{
+namespace mixin
+{
+
+//------------------------------------------------------------------------------
+namespace input
+{
+
+// TODO it'd be nice to specify a lambda expression that'd replace the call to
+// thrust::sequence when desired.
```
brycelelbach, line 247: The idea here is to parameterize initialization?

testing/async/mixin.h
```diff
@@ -0,0 +1,645 @@
...
+    thrust::sequence(input.begin(),
+                     input.end(),
+                     static_cast<value_type>(1.5),
+                     static_cast<value_type>(1));
+    return input;
+  }
+};
+
+template <typename value_type>
+struct counting_iterator_0
```
brycelelbach, line 265: Could this be `_starts_at_0` or `_from_0`, etc?

testing/async/mixin.h
```diff
@@ -0,0 +1,645 @@
...
+  }
+};
+
+struct discard_iterator
+{
+  struct output_type
+  {
+    using iterator = thrust::discard_iterator<>;
+
+    iterator begin() { return thrust::make_discard_iterator(); }
```
brycelelbach, line 386: I don't think we need the non-const here.

testing/async/mixin.h
```diff
@@ -0,0 +1,645 @@
...
+
+  template <typename InputType>
+  static output_type generate_output(std::size_t /*num_values*/,
+                                     InputType& input)
+  {
+    return input;
+  }
+};
+
+struct discard_iterator
```
brycelelbach, line 380: @griwes suggests calling these `*_range`.

testing/async/mixin.h
```diff
@@ -0,0 +1,645 @@
...
+template <typename value_type>
+struct constant_iterator_1
+{
+  struct input_type
+  {
+    using iterator = thrust::constant_iterator<value_type>;
+
+    std::size_t num_values;
+
+    iterator begin() { return iterator{static_cast<value_type>(1)}; }
```
brycelelbach, line 325: I don't think we need the non-const here.

testing/async/mixin.h
```diff
@@ -0,0 +1,645 @@
...
+template <typename value_type>
+struct counting_iterator_1
+{
+  struct input_type
+  {
+    using iterator = thrust::counting_iterator<value_type>;
+
+    std::size_t num_values;
+
+    iterator begin() { return iterator{static_cast<value_type>(1)}; }
```
brycelelbach, line 299: I don't think we need the non-const here.

testing/async/mixin.h
```diff
@@ -0,0 +1,645 @@
...
+    iterator begin() const { return iterator{static_cast<value_type>(1)}; }
+    iterator cbegin() const { return iterator{static_cast<value_type>(1)}; }
+
+    iterator end() { return iterator{static_cast<value_type>(1 + num_values)}; }
```
brycelelbach, line 303: I don't think we need the non-const here.

testing/async/mixin.h
```diff
@@ -0,0 +1,645 @@
...
+  static input_type generate_input(std::size_t num_values)
+  {
+    return {num_values};
+  }
+};
+
+template <typename value_type>
+struct constant_iterator_1
+{
+  struct input_type
```
brycelelbach, line 319: We should probably explain somewhere that the `input_type`s need random access (thus begin, end, and a size), while the `output_type`s (like discard) only need to satisfy output iterator / range.

testing/async/mixin.h
```diff
@@ -0,0 +1,645 @@
...
+namespace postfix_args
+{
+/* Defined per algorithm. Example:
+ *
+ * // Defines several overloads:
+ * // algorithm([policy,] input, output) // no postfix args
+ * // algorithm([policy,] input, output, initial_value)
+ * // algorithm([policy,] input, output, initial_value, binary_op)
+ * template <typename value_type, typename alternate_binary_op =
+ * thrust::maximum<>> struct all_overloads
```
brycelelbach, line 411: Bad line wrapping.

testing/async/mixin.h
```diff
@@ -0,0 +1,645 @@
...
+ * {
+ *   using postfix_args_type = std::tuple<         // List any extra arg
+ * overloads: std::tuple<>,                               // - no extra args
```
brycelelbach, line 414: Bad line wrapping.

testing/async/mixin.h
```diff
@@ -0,0 +1,645 @@
...
+//------------------------------------------------------------------------------
+namespace compare_outputs
+{
+
+namespace detail
+{
+
+void basic_event_validation(thrust::device_event& e)
+{
+  TEST_EVENT_WAIT(e);
```
brycelelbach, line 521: Do we still need the macros if we're just abstracting this into a function? Can we nuke the macros?

testing/async/mixin.h
```diff
@@ -0,0 +1,645 @@
...
+                              OutputType const& test)
+  {
+    detail::basic_event_validation(e);
+    ASSERT_EQUAL(ref, test);
+  }
+};
+
+// Does an 'almost_equal' comparison for floating point types, since fp
+// addition is non-associative
+struct assert_maybe_fuzzy_equal
```
brycelelbach, line 546: Can we call this `assert_almost_equal`?

testing/async/mixin.h
```diff
@@ -0,0 +1,645 @@
...
+{
+private:
+  template <typename EventType, typename OutputType>
+  static void compare_outputs_impl(EventType& e,
+                                   OutputType const& ref,
+                                   OutputType const& test,
+                                   std::false_type /* is_floating_point */)
+  {
+    detail::basic_event_validation(e);
+    ASSERT_EQUAL(ref, test);
```
brycelelbach, line 556: Can we use `ASSERT_ALMOST_EQUAL` for integer types?

testing/async/mixin.h
```diff
@@ -0,0 +1,645 @@
...
+                              OutputType const& test)
+  {
+    detail::basic_event_validation(e);
+    ASSERT_EQUAL_QUIET(ref, test);
+  }
+};
+
+// Does an 'almost_equal' comparison for floating point types, since fp
+// addition is non-associative
+struct assert_maybe_fuzzy_equal_quiet
```
brycelelbach, line 594: s/maybe_fuzzy/almost/

testing/async/test_policy_overloads.h
```diff
@@ -0,0 +1,421 @@
...
+      num_values);
+  }
+
+  // Wrap the basic_policy_test in a try/catch block to inject helpful info into
+  // any thrown unit test exceptions.
+  template <std::size_t postfix_idx, typename PolicyTuple, typename ValidateEvent>
+  static void run_basic_policy_test(std::string const &policy_desc,
+                                    PolicyTuple &&policy_tuple,
+                                    ValidateEvent &&validate,
+                                    std::size_t const num_values)
```
brycelelbach, line 155: I have slight preference for this being combined with the other `run_basic_policy_test` overload, or call one of them `_impl`; they don't do the same thing, so they should not be overloads.

brycelelbach:
Few high-level remarks:

* I'd like this to support input initialization on the host because there are some initialization strategies that are host only, like `unittest::random_integer`.
* I'd prefer to only generate input once and then copy.
* I'd like to retain the `unittest::random_integer` tests for async algorithms.

testing/async/test_policy_overloads.h
```diff
@@ -0,0 +1,421 @@
...
+  static void test_postfix_overloads(std::size_t const,
+                                     size_const<num_postfix_arg_sets>)
+  {
+    // terminal case, no-op
+  }
+
+  //----------------------------------------------------------------------------
+  // For the specified postfix overload set, test the algorithm with several
+  // different policy configurations.
+  template <std::size_t postfix_idx>
```
brycelelbach, line 101: Template parameters should be `CamelCase`.

testing/async/test_policy_overloads.h
```diff
@@ -0,0 +1,421 @@
...
+    validate(e_a);
+    validate(e_b);
+    validate(e_c);
+    validate(e_d);
+  }
+
+  //----------------------------------------------------------------------------
+  // Test .after(event/future) handling:
+  template <std::size_t postfix_idx>
+  static void run_after_future_test(std::size_t const num_values)
```
brycelelbach, line 272: Let's try to unify the naming scheme. We have:
* Two overloads for `run_basic_policy_test`, one which has the try block, one which doesn't. Then there's also `basic_policy_test`.
* `run_after_future_test` and `after_future_test`, which appear to have the same relationship as the `run_basic_policy_test` overloads.

testing/async/test_policy_overloads.h
```diff
@@ -0,0 +1,421 @@
...
+                                     size_const<num_postfix_arg_sets>)
+  {
+    // terminal case, no-op
+  }
+
+  //----------------------------------------------------------------------------
+  // For the specified postfix overload set, test the algorithm with several
+  // different policy configurations.
+  template <std::size_t postfix_idx>
+  static void run_basic_policy_tests(std::size_t const num_values)
```
brycelelbach, line 102: Let's just call this one `basic_policy_test` to mirror `after_future_test`.

testing/async/test_policy_overloads.h
```diff
@@ -0,0 +1,421 @@
...
+  }
+
+  template <std::size_t postfix_idx>
+  static void after_future_test(std::size_t const num_values)
+  {
+    using postfix_tuple_type =
+    std::tuple_element_t<postfix_idx, postfix_args_type>;
+    postfix_tuple_type const postfix_tuple = get_postfix_tuple<postfix_idx>();
+
+    // Generate index sequences for the tuples. Prefix size always = 1 here.
```
brycelelbach, line 309: Can we expand on this comment here explaining why prefix size is always 1?

testing/async/test_policy_overloads.h
```diff
@@ -0,0 +1,421 @@
...
+                          input_tmp,
+                          output_tmp,
+                          postfix_tuple,
+                          postfix_index_seq{});
+                        THRUST_UNUSED_VAR(x),
+                        thrust::event_error,
+                        thrust::event_error(thrust::event_errc::no_state));
+
+    // Explicitly order this invocation after e_b:
+    auto policy_after_e_b = thrust::device.after(e_b);
```
brycelelbach, line 367: We should leave a comment here explaining that it is important that in the first of these tests, the execution policy is an rvalue and in the second of these tests it is an lvalue; both are supposed to be consumable.

testing/async/test_policy_overloads.h
```diff
@@ -0,0 +1,421 @@
...
+    // Make sure the tuple contains an lvalue ref:
```
brycelelbach, line 368: Can we say something like "we can't use make_tuple because it would make a copy".

testing/async/test_policy_overloads.h
```diff
@@ -0,0 +1,421 @@
...
+                             output_c,
+                             postfix_tuple,
+                             postfix_index_seq{});
+    ASSERT_EQUAL(true, e_c.valid_stream());
+    auto const stream_c = e_c.stream().native_handle();
+
+    // Should use same stream as e_b:
+    ASSERT_EQUAL_QUIET(stream_b, stream_c);
+
+    // Verify that indirect double consumption of e_b produces an exception:
```
brycelelbach, line 383: s/indirect/lvalue/

testing/async/test_policy_overloads.h
```diff
@@ -0,0 +1,421 @@
...
+                             output_b,
+                             postfix_tuple,
+                             postfix_index_seq{});
+    ASSERT_EQUAL(true, e_b.valid_stream());
+    auto const stream_b = e_b.stream().native_handle();
+
+    // Second invocation should use same stream as before:
+    ASSERT_EQUAL_QUIET(stream_a, stream_b);
+
+    // Verify that double consumption of e_a produces an exception:
```
brycelelbach, line 354: Say "rvalue double consumption" to be consistent with below.

thrust/system/cuda/detail/async/exclusive_scan.h
```diff
@@ -0,0 +1,199 @@
+/******************************************************************************
+ * Copyright (c) 2016, NVIDIA CORPORATION.  All rights reserved.
```
brycelelbach, line 2: Adjust copyright dates.

thrust/system/cuda/detail/async/exclusive_scan.h
```diff
@@ -0,0 +1,199 @@
...
+          typename OutputIt,
+          typename InitialValueType,
+          typename BinaryOp>
+auto async_exclusive_scan_n(execution_policy<DerivedPolicy>& policy,
+                            ForwardIt first,
+                            Size n,
+                            OutputIt out,
+                            InitialValueType init,
+                            BinaryOp op)
+  -> unique_eager_event
```
brycelelbach, line 76: Don't use a trailing return type that doesn't depend on anything.

thrust/system/cuda/detail/async/exclusive_scan.h
```diff
@@ -0,0 +1,199 @@
...
+  using Dispatch32 = cub::DispatchScan<ForwardIt,
+                                       OutputIt,
+                                       BinaryOp,
+                                       InitialValueType,
+                                       thrust::detail::int32_t>;
+  using Dispatch64 = cub::DispatchScan<ForwardIt,
+                                       OutputIt,
+                                       BinaryOp,
+                                       InitialValueType,
+                                       thrust::detail::int64_t>;
```
brycelelbach, line 87: Ouch, I don't think `async::reduce` does dispatch correctly; can we either do this in this patch, or file an issue to track that?

thrust/system/cuda/detail/async/inclusive_scan.h
```diff
@@ -0,0 +1,195 @@
...
+                            OutputIt out,
+                            BinaryOp op)
+  -> unique_eager_event
+{
+  using Dispatch32 = cub::DispatchScan<ForwardIt,
+                                       OutputIt,
+                                       BinaryOp,
+                                       cub::NullType,
+                                       thrust::detail::int32_t>;
+
```
brycelelbach, line 81: There was no newline between these two `usings` in `exclusive_scan`.

thrust/async/scan.h
```diff
@@ -0,0 +1,355 @@
+/*
```
brycelelbach, line 1: We should have separate headers for each of the algorithms.

thrust/async/scan.h
```diff
@@ -0,0 +1,355 @@
...
+            typename OutputIt,
+            typename BinaryOp>
+  THRUST_NODISCARD
+  auto
+  operator()(thrust::detail::execution_policy_base<DerivedPolicy> const& exec,
+             ForwardIt&& first,
+             Sentinel&& last,
+             OutputIt&& out,
+             BinaryOp&& op) const
+    // ADL dispatch.
```
brycelelbach, line 113: Extra spaces before `//`.

testing/async/inclusive_scan/using_vs_adl.cu
```diff
@@ -0,0 +1,169 @@
...
+
+    using OutIter = thrust::remove_cvref_t<decltype(host_output.begin())>;
+
+    // ADL should resolve this to the synchronous `thrust::` algorithm.
+    // This is checked by ensuring that the call returns an output iterator.
+    OutIter result =
+      inclusive_scan(host_input.cbegin(),
+                     host_input.cend(),
+                     host_output.begin(),
+                     std::get<PostfixArgIndices>(THRUST_FWD(postfix_tuple))...);
```
brycelelbach, line 45: I don't think synchronous scans are marked nodiscard, so I think we can just ignore the result altogther.

testing/async/inclusive_scan/large_types.cu
```diff
@@ -0,0 +1,63 @@
...
+
+struct TestLargeTypes
+{
+  void operator()(std::size_t num_values) const
+  {
+    using testing::async::test_policy_overloads;
+
+    test_policy_overloads<invoker<FixedVector<int, 1>>>::run(num_values);
+
+#ifndef __QNX__ // These are excluded on QNX for scan.cu's version of this test.
```
brycelelbach, line 52: Take this out, it should no longer be a problem on QNX. Maybe file an issue to take this out from the synchronous scan tests too?

testing/async/inclusive_scan/mixed_types.cu
```diff
@@ -0,0 +1,95 @@
...
+  }
+};
+
+} // namespace
+
+void TestScanMixedTypes()
+{
+  // 10 values are enough to check the behavior we want to test while staying
+  // small enough to reason about during debugging.
+  constexpr std::size_t num_values = 10;
```
brycelelbach, line 84: This is incorrect, because CUB has different codepaths for single tile versus multi-tile; we need to test both, so this should use the testing mechanism that supports a variable problem size (ditto for the corresponding exclusive_scan test).

hcedwar:
The "files changed" code volume is quite large.  Where is an algorithm & design summary to review this aspect vs. code details?

allisonvacanti:
They're mostly changes to Thrust's internal testing infrastructure. The actual API changes here are very small and follow the existing conventions and design plan for Thrust's existing async algorithms.

There's a design doc from a couple years ago floating around for the async design. If you need to see it for some reason I can dig it up.

testing/async/mixin.h
allisonvacanti: Updated to `assert_almost_equal_if_fp`.

testing/async/mixin.h
allisonvacanti: I avoid that because it's not necessary here -- these are for the scan algorithms which give exact results for integer types but approximate results for fp due to non-associative addition. So checking with exactly equal for integers is intentional here, see the comment on the struct containing this implementation.

testing/async/mixin.h
allisonvacanti: Updated as above.

testing/async/test_policy_overloads.h
allisonvacanti: These aren't overloads -- `run_basic_policy_tests` (plural) launches tests for each policy configurations by calling `run_basic_policy_test` (singular).

But I do like your idea of doing a function-scope try block in `basic_policy_test` instead of having the `run_basic_policy_test` trampoline, so I'll do that.

allisonvacanti:
We discussed this in a call, but to update the PR review:

> * I'd like this to support input initialization on the host because there are some initialization strategies that are host only, like `unittest::random_integer`.

The input can be initialized on the host and copied to the device if a host-only strategy is needed. Since the input is more frequently accessed on the device (4 device invocations vs 1 host invocation), this reduces the amount of transfer overhead in the common case that inputs can be generated on the device.

> * I'd prefer to only generate input once and then copy.

Me too, but I'd also like to not require copyability for the generic `input_type` concept. If this ends up being problematic in practice I'm happy to revisit this.

> * I'd like to retain the `unittest::random_integer` tests for async algorithms.

Can do. I'll see about adapting it to run on the device, too -- we have device-side rng facilities, so hopefully it'll be trivial.

testing/async/test_policy_overloads.h
allisonvacanti: Done.

testing/async/test_policy_overloads.h
allisonvacanti: Renamed so that both tests have a `run_XXX_tests` entry point, which can launch multiple `XXX_test` implementations when needed.

testing/async/test_policy_overloads.h
allisonvacanti: Done.

testing/async/test_policy_overloads.h
allisonvacanti: Clarified value categories in comments.

thrust/system/cuda/detail/async/exclusive_scan.h
allisonvacanti: This will be fixed with NVIDIA/cub#212.

thrust/async/scan.h
allisonvacanti: I'm going to hold off on this -- we should think about more granular headers in general in thrust. Since the `thrust::` scan algorithms are all in a single public header, I'd rather the async public headers match those conventions.

testing/async/inclusive_scan/using_vs_adl.cu
allisonvacanti: Good point, I got carried away following the reduce algorithms as a reference :) I've removed the nodiscard annotations completely.

testing/async/inclusive_scan/large_types.cu
allisonvacanti: #1345

testing/async/inclusive_scan/mixed_types.cu
allisonvacanti: I added a `DECLARE_SIZED_UNITTEST` macro and used it for these. The new macro was needed since the others all expect the test to take template parameters.

allisonvacanti:
Addressed feedback. The `mixed_types` tests are failing now, I'll need to investigate this. We also discussed shortening the namespace for some of the device-specific test mixins, I'll fix that when I take a look at the new test failures.

allisonvacanti:
`mixed_types` test should still be failing, but running DVS to try to repro build timeouts.

DVS CL: 29371528

allisonvacanti:
Latest round of DVS runs show the following issues on Clang:

```
thrust/system/cuda/detail/async/exclusive_scan.h:160:8: error: moving a local object in a return statement prevents copy elision [-Werror,-Wpessimizing-move]
thrust/system/cuda/detail/async/inclusive_scan.h:158:8: error: moving a local object in a return statement prevents copy elision [-Werror,-Wpessimizing-move]
```

testing/async/exclusive_scan/basic.cu
allisonvacanti: 1. Changed to `snake_case` for the all new async test ✔
2. `exclusive_scan` appears in the executable and test name. These are subtests of `exclusive_scan`, including the algorithm name for each subtest here would be redundant and verbose.
3. Changed `basic` to `simple`.

testing/async/exclusive_scan/large_indices.cu
allisonvacanti: Done.

testing/async/exclusive_scan/large_indices.cu
allisonvacanti: Added a note.

testing/async/exclusive_scan/large_indices.cu
allisonvacanti: Added a note.

testing/async/exclusive_scan/large_indices.cu
allisonvacanti: These iterator implementations are unique enough that we need different ones for the different scans.

I may try to abstract this later as we expand the new-style tests to more algorithms, we'll see how that goes.

testing/async/exclusive_scan/large_types.cu
allisonvacanti: Done.

testing/async/mixin.h
allisonvacanti: Turns out this was just a typo in the docs -- the actual implementation only uses a single `mixin` in the namespace path.

testing/async/test_policy_overloads.h
allisonvacanti: The async framework is CUDA-only for the foreseeable future. If it gets expanded to other backends we can address this at that time.

testing/async/test_policy_overloads.h
allisonvacanti: Done.

thrust/system/cuda/detail/async/scan.h
allisonvacanti: This matches the existing convention used for the sync scans and simplifies the ADL dispatch headers a bit. Let's consider the "separate algorithms per header" as an implementation detail of the CUDA backend here. I think it'd be nice to eventually split the headers out for both sync/async scans at some point as a cleanup.

testing/async/mixin.h
allisonvacanti: Done.

testing/async/mixin.h
allisonvacanti: Yep, it'd be a nice feature to have at some point, but there are easy workaround for now.

testing/async/mixin.h
allisonvacanti: Changed to `_from_0`.

testing/async/mixin.h
allisonvacanti: Removed.

testing/async/mixin.h
allisonvacanti: The current name matches the name of what we're testing. I prefer this as-is.

It's scoped to a namespace so we can always differentiate this from the thrust class.

testing/async/mixin.h
allisonvacanti: Added this info to the appropriate sections in the class documentation.

testing/async/mixin.h
allisonvacanti: Fixed.

testing/async/mixin.h
allisonvacanti: That sounds like a great candidate for a follow-up PR.

allisonvacanti:
run tests

brycelelbach:
run tests

allisonvacanti:
run tests

allisonvacanti:
run tests

allisonvacanti:
DVS CL: 29473661

brycelelbach (approved):

PR #1253, opened 2020-08-06
Restore some THRUST_DECLTYPE_RETURNS macros in async test implementations

### LINKED ISSUE #1250: Some async algorithm tests fail on C++17
It looks like 7ff227ae12a927ba9aa62f216f658c939d21785f broke some async algorithms on C++17 with gcc 7.5.0.

For example, `thrust.cpp.cuda.cpp17.test.async_reduce_into` is failing with an "invalid device function" error.

allisonvacanti:
This partially reverts 7ff227ae12a927ba9aa62f216f658c939d21785 and
fixes #1250.

I'm not sure why changing these broke the tests, but since these
usages are just testing details that are being refactored by #1251
let's just revert the change for now.

The test failures were only happening on GCC, MSVC was fine with both
versions of these functions, so it may be a compiler issue.

### DIFF
--- a/testing/async_reduce.cu
+++ b/testing/async_reduce.cu
@@ -48,7 +48,7 @@ struct custom_plus
     auto operator()(                                                          \
       ForwardIt&& first, Sentinel&& last                                      \
     )                                                                         \
-    THRUST_RETURNS(                                                           \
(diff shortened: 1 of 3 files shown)


allisonvacanti:
DVS CL 28912962

PR #1255, opened 2020-08-13
reduce cudaDeviceSynchronize calls

rongou:
While profiling RAPIDS cuDF applications, I saw a lot of `cudaDeviceSynchronize` calls, which may be detrimental to multi-threaded clients using per-thread default stream. It seems to be a bug to be making these calls. This PR simply moves the logic in `execute_on_stream_base` to the base entry point so that all callers can benefit. With this change I no longer see `cudaDeviceSynchronize` calls in the application I'm profiling.

All tests passed locally.

@brycelelbach @allisonvacanti

### DIFF
--- a/thrust/system/cuda/detail/par.h
+++ b/thrust/system/cuda/detail/par.h
@@ -69,31 +69,6 @@ private:
   {
     return exec.stream;
   }
-
-  friend __host__ __device__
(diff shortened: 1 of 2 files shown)


allisonvacanti:
Launched CI on 28935982.

PR #1257, opened 2020-08-14
Updated shuffle implementation to use better hash function

djns99:
Fixes https://github.com/thrust/thrust/issues/1256 by changing the hash function from taus88 to use wyhash. Adds test for a random distribution of numbers

### DIFF
--- a/testing/shuffle.cu
+++ b/testing/shuffle.cu
@@ -1,15 +1,309 @@
 #include <thrust/detail/config.h>
 
 #if THRUST_CPP_DIALECT >= 2011
+#include <map>
+#include <limits>
 #include <thrust/random.h>
 #include <thrust/sequence.h>
 #include <thrust/shuffle.h>
-#include <thrust/sort.h>
 #include <unittest/unittest.h>
-#include <map>
 
-template <typename Vector>
-void TestShuffleSimple() {
+class CephesFunctions {
+public:
+  static double cephes_igamc(double a, double x) {
+    double ans, ax, c, yc, r, t, y, z;
+    double pk, pkm1, pkm2, qk, qkm1, qkm2;
+
+    if ((x <= 0) || (a <= 0))
+      return (1.0);
+
+    if ((x < 1.0) || (x < a))
+      return (1.e0 - cephes_igam(a, x));
+
+    ax = a * log(x) - x - cephes_lgam(a);
+
+    if (ax < -MAXLOG) {
+      printf("igamc: UNDERFLOW\n");
+      return 0.0;
+    }
+    ax = exp(ax);
+
+    /* continued fraction */
+    y = 1.0 - a;
+    z = x + y + 1.0;
+    c = 0.0;
+    pkm2 = 1.0;
+    qkm2 = x;
+    pkm1 = x + 1.0;
+    qkm1 = z * x;
+    ans = pkm1 / qkm1;
+
+    do {
+      c += 1.0;
+      y += 1.0;
+      z += 2.0;
+      yc = y * c;
+      pk = pkm1 * z - pkm2 * yc;
+      qk = qkm1 * z - qkm2 * yc;
+      if (qk != 0) {
+        r = pk / qk;
+        t = fabs((ans - r) / r);
+        ans = r;
+      } else
+        t = 1.0;
+      pkm2 = pkm1;
+      pkm1 = pk;
+      qkm2 = qkm1;
+      qkm1 = qk;
+      if (fabs(pk) > big) {
+        pkm2 *= biginv;
+        pkm1 *= biginv;
+        qkm2 *= biginv;
+        qkm1 *= biginv;
+      }
+    } while (t > MACHEP);
+
+    return ans * ax;
+  }
+
+private:
+  static constexpr double rel_error = 1E-12;
+
+  static constexpr double MACHEP = 1.11022302462515654042E-16;  // 2**-53
+  static constexpr double MAXLOG = 7.09782712893383996732224E2; // log(MAXNUM)
+  static constexpr double MAXNUM = 1.7976931348623158E308; // 2**1024*(1-MACHEP)
+  static constexpr double PI = 3.14159265358979323846;
+
+  static constexpr double big = 4.503599627370496e15;
+  static constexpr double biginv = 2.22044604925031308085e-16;
+
+  static int sgngam;
+
+  static double cephes_igam(double a, double x) {
+    double ans, ax, c, r;
+
+    if ((x <= 0) || (a <= 0))
+      return 0.0;
+
+    if ((x > 1.0) && (x > a))
+      return 1.e0 - cephes_igamc(a, x);
+
+    /* Compute  x**a * exp(-x) / gamma(a)  */
+    ax = a * log(x) - x - cephes_lgam(a);
+    if (ax < -MAXLOG) {
+      printf("igam: UNDERFLOW\n");
+      return 0.0;
+    }
+    ax = exp(ax);
+
+    /* power series */
+    r = a;
+    c = 1.0;
+    ans = 1.0;
+
+    do {
+      r += 1.0;
+      c *= x / r;
+      ans += c;
+    } while (c / ans > MACHEP);
+
+    return ans * ax / a;
+  }
+
+  /* A[]: Stirling's formula expansion of log gamma
+   * B[], C[]: log gamma function between 2 and 3
+   */
+  static constexpr unsigned short A[] = {
+      0x6661, 0x2733, 0x9850, 0x3f4a, 0xe943, 0xb580, 0x7fbd,
+      0xbf43, 0x5ebb, 0x20dc, 0x019f, 0x3f4a, 0xa5a1, 0x16b0,
+      0xc16c, 0xbf66, 0x554b, 0x5555, 0x5555, 0x3fb5};
+  static constexpr unsigned short B[] = {
+      0x6761, 0x8ff3, 0x8901, 0xc095, 0xb93e, 0x355b, 0xf234, 0xc0e2,
+      0x89e5, 0xf890, 0x3d73, 0xc114, 0xdb51, 0xf994, 0xbc82, 0xc131,
+      0xf20b, 0x0219, 0x4589, 0xc13a, 0x055e, 0x5418, 0x0c67, 0xc12a};
+  static constexpr unsigned short C[] = {
+      /*0x0000,0x0000,0x0000,0x3ff0,*/
+      0x12b2, 0x1cf3, 0xfd0d, 0xc075, 0xd757, 0x7b89, 0xaa0d, 0xc0d0,
+      0x4c9b, 0xb974, 0xeb84, 0xc10a, 0x0043, 0x7195, 0x6286, 0xc131,
+      0xf34c, 0x892f, 0x5255, 0xc143, 0xe14a, 0x6a11, 0xce4b, 0xc13e};
+
+  static constexpr double MAXLGM = 2.556348e305;
+
+  /* Logarithm of gamma function */
+  static double cephes_lgam(double x) {
+    double p, q, u, w, z;
+    int i;
+
+    sgngam = 1;
+
+    if (x < -34.0) {
+      q = -x;
+      w = cephes_lgam(q); /* note this modifies sgngam! */
+      p = floor(q);
+      if (p == q) {
+      lgsing:
+        goto loverf;
+      }
+      i = (int)p;
+      if ((i & 1) == 0)
+        sgngam = -1;
+      else
+        sgngam = 1;
+      z = q - p;
+      if (z > 0.5) {
+        p += 1.0;
+        z = p - q;
+      }
+      z = q * sin(PI * z);
+      if (z == 0.0)
+        goto lgsing;
+      /*      z = log(PI) - log( z ) - w;*/
+      z = log(PI) - log(z) - w;
+      return z;
+    }
+
+    if (x < 13.0) {
+      z = 1.0;
+      p = 0.0;
+      u = x;
+      while (u >= 3.0) {
+        p -= 1.0;
+        u = x + p;
+        z *= u;
+      }
+      while (u < 2.0) {
+        if (u == 0.0)
+          goto lgsing;
+        z /= u;
+        p += 1.0;
+        u = x + p;
+      }
+      if (z < 0.0) {
+        sgngam = -1;
+        z = -z;
+      } else
+        sgngam = 1;
+      if (u == 2.0)
+        return (log(z));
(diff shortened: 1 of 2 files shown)


thrust/system/detail/generic/shuffle.inl
```diff
@@ -32,6 +32,44 @@ namespace system {
...
 
+class round_function {
+private:
+  constexpr static uint64_t _wyp0 = 0xa0761d6478bd642full;
+  constexpr static uint64_t _wyp1 = 0xe7037ed1a0b428dbull;
+  constexpr static uint64_t _wyp2 = 0x8ebc6af09c88c6e3ull;
+  constexpr static uint64_t _wyp3 = 0x589965cc75374cc3ull;
+  constexpr static uint64_t _wyp4 = 0x1d8e4e27c47d124full;
+
+#ifdef __CUDA_ARCH__
```
djns99, line 43: Is this the correct method of switching behaviour for host vs device implementations

testing/shuffle.cu
```diff
@@ -139,4 +135,53 @@ void TestShuffleUniformPermutation() {
   ASSERT_LESS(chi_squared, confidence_threshold);
 }
 DECLARE_VECTOR_UNITTEST(TestShuffleUniformPermutation);
+
+template <typename Vector> void TestShuffleEvenSpacingBetweenOccurances() {
+  typedef typename Vector::value_type T;
+  const uint64_t shuffle_size = 10;
```
RAMitchell, line 141: The bug report only had problems on sequences > 150, let's make sure we catch it with  this test.

Is there a way to set up this test was similar to work on varying sizes? I think there is a test macro for inputting various size arguments.

We might have to implement a helper function here for the chi squared threshold.

thrust/system/detail/generic/shuffle.inl
```diff
@@ -32,6 +32,44 @@ namespace system {
 namespace detail {
 namespace generic {
 
+class round_function {
```
RAMitchell, line 35: I don't like using something like wyhash without a reference. I would prefer using some xor combination of internal thrust random engines from here: https://thrust.github.io/doc/group__predefined__random.html. These engines all have academic references, and we can also reference numerical recipes which recommends combining different types of engines as best practice.

The downside is that this might be a bit slower, but it could be more justified.

thrust/system/detail/generic/shuffle.inl
```diff
@@ -119,7 +147,7 @@ class feistel_bijection {
   uint64_t left_side_bits;
   uint64_t right_side_mask;
   uint64_t left_side_mask;
-  uint64_t key[num_rounds];
+  uint64_t key[num_rounds * 2];
```
RAMitchell, line 150: Where does this 2 come from?

testing/shuffle.cu
djns99: No the bug report still had issues with smaller, they just werent as immediately obvious. For example you will notice every 5th or 6th number in the 100 length one is very close together

djns99: We can do larger sizes but this has a factor of n^2 so we would have to significantly reduce the sample size.

thrust/system/detail/generic/shuffle.inl
djns99: Just because the wyhash function I used took two keys. In my original testing I found that you needed at least 128 bits of randomness to consistently random results

thrust/system/detail/generic/shuffle.inl
djns99: Fair enough, the difficulty here is actually generating sufficient diffusion. The important feature of this function is that one bit flip causes a large change in the output. This is a feature of hash functions, not random number generators. Perhaps a more well known hash function such as crc would be acceptable?

djns99: Perhaps https://blake2.net/

djns99: Additionally, many of the good random number generators are expensive to construct since they are intended to be create once call many

djns99: I tested crc32c but it performed very poorly, which is to be expected somewhat since it is a checksum not a hash function. Additionally it is only 32 bit

djns99: Ok I did some brief research
https://doi.org/10.1002/spe.2179 and concludes that MurmurHash2 has good distribution - particularly the section on avalanche affect.
Additionally, https://softwareengineering.stackexchange.com/questions/49550/which-hashing-algorithm-is-best-for-uniqueness-and-speed provides some benchmarks for a number of functions and also suggests (albeit much less scientifically) that MurmurHash2 has strong randomness.
It seems reasonably fast since we dont need to worry about the overhead of being cryptographically secure and we dont particularly care about collision resistance properties, unless they are very common for close keys.

djns99: I have pushed a commit with the changes to MurmurHash2. I believe this is slightly weaker than WyHashv4 but it is still much stronger than a random number generator would be. Currently both fail the tests due to insufficient randomness. To fix them we may need to increase the number of rounds, as well as use a stronger random number generator as input

RAMitchell: > The important feature of this function is that one bit flip causes a large change in the output. This is a feature of hash functions, not random number generators.

Not sure if that's true, consecutive seeds for random number generators should absolutely be uncorrelated. I think the key difference for the hash functions is that they are designed for larger inputs. For our purposes 64 -> 64 bits is fine, so I'm not convinced we need anything other than a good random number generator.

testing/shuffle.cu
RAMitchell: Got it. I wonder if we could come up with something linear time, not necessarily for this PR. Then we could effectively fuzz test a large range of inputs.

Another thought I had, we could test the feistel is invertible to catch any potential issues there.

thrust/system/detail/generic/shuffle.inl
RAMitchell: Definitely increase the number of rounds though, even if we have to go up to 16+.

djns99: >Not sure if that's true, consecutive seeds for random number generators should absolutely be uncorrelated. I think the key difference for the hash functions is that they are designed for larger inputs. For our purposes 64 -> 64 bits is fine, so I'm not convinced we need anything other than a good random number generator.

This maybe the case for cryptographic DRBGs, but not necessarily for general random number generator. The diffusion must come from somewhere, either through a very expensive initialisation procedure such as for mersenne twister, which may possibly involve running the prng many times. Alternatively, the random number generator may use a hash function to add diffusion, but then you might as well just use the hash function.
The difference here is that a random number generator such as a LFSR will diverge quickly but will start out very similar if the register is initialised similar, but will still output a similar sequence for the first few values, which is especially concerning since we only ever use the first 8 bits generated for an array of length 256.
One thing to note with the previous implementation is that since the key was applied after the PRNG, the output would always be the same for the item with the corresponding index at a given round. This could potentially lead to previous XOR's being reversed causing each value to only be XOR'd with the key, limiting the total randomness

testing/shuffle.cu
djns99: That would be a good test for the feistel, all you would have to do is test that no items are duplicated/missed and they are all within range.

As for a linear time test I can grab the test for number of times each item occurs at a given location

allisonvacanti:
@djns99 @RAMitchell Just checking in -- it looks like you two have been iterating on this PR, just let me know when you're ready for me to do a final review and start testing/landing it.

RAMitchell:
@djns99 I'm happy with your current approach using murmurhash, did you have anything else you wanted to add? Would be good to have a couple of comments with references for CephesFunctions and murmerhash. Please also check the licenses are compatible i.e. we are allowed to use their code in this way.

djns99:
> @djns99 I'm happy with your current approach using murmurhash, did you have anything else you wanted to add? Would be good to have a couple of comments with references for CephesFunctions and murmerhash. Please also check the licenses are compatible i.e. we are allowed to use their code in this way.

I probably wont have time to do this in the next few days. However, I am still having difficulty obtaining sufficiently random results, even if I increase the number of rounds. Some preliminary investigations indicate that it is likely the murmur hash function with insufficient randomness, however it is a significant improvement over the current results even at 8 rounds. 
I would also note that it could be a test bug since I would expect a high enough number of rounds to pass.


As for licensing, cephes was taken from the NIST suite for testing randomness, which according to https://csrc.nist.gov/Projects/Random-Bit-Generation/Documentation-and-Software is in the public domain. I took murmur hash from the smhasher repo, the licensing information there say it is also in the public domain.

RAMitchell:
Ok, I think the key thing is for us to make the next toolkit release (11.1?). Let's catch up next week on this.

djns99:
@RAMitchell I have done a bit of work on this and decided to use the RC5 encryption algorithm https://doi.org/10.1007/3-540-60590-8_7. This uses a 12-20 round (I have it set to 12 at present) feistel network using data dependent rotations on an w-bit register. The only shortfall of this is that it requires us to round to a power of 4 instead of 2. However, it means we can make some guarantees about the strength of our internal function (i.e. any errors are due to an insufficiently random key).
There are still some issues with the test I am in the process of diagnosing

As for licensing, the patent for the algorithm expired in 2015 https://patents.google.com/patent/US5835600A/en so I believe this means we are allowed to use it.

djns99:
Test issues are resolved.

thrust/system/detail/generic/shuffle.inl
```diff
@@ -32,94 +32,90 @@ namespace system {
...
-    return (value_hash ^ key) & left_side_mask;
+  __host__ __device__ uint32_t get_mask() const
+  {
+    return (uint32_t)((1ull << (uint64_t)w) - 1ull);
+  }
+
+  __host__ __device__ uint32_t rotl( uint32_t val, uint32_t amount ) const
+  {
+    assert((val & ~get_mask()) == 0);
+    const uint32_t amount_mod = w_is_pow_2 ? (amount & (w-1)) : (amount % w);
```
RAMitchell, line 100: Is this a performance optimisation? If so, I would just use %. GPU code does not like branches. If the compiler uses predication, both branches get executed always.

RAMitchell (commented):
Regarding the use of RC5, that's up to the thrust team. You might want to have an alternative in mind.

> The only shortfall of this is that it requires us to round to a power of 4 instead of 2.

I'm a bit confused, current code still uses nearest power of two in `get_cipher_bits`.

thrust/system/detail/generic/shuffle.inl
djns99: GPU code doesnt like branches if the warp/half warp diverges. In this case it will always take one branch. The branch predictor should also be able to predict the correct branch since again - always takes the same.
My understanding is that while GPUs dont like branches they like modulus/division even less.

djns99:
> I'm a bit confused, current code still uses nearest power of two in `get_cipher_bits`.

Note the change to 

```
m >>= 2u;
```

This counts the number of pairs of bits now, with `w` representing one half of the feistel block

thrust/system/detail/generic/shuffle.inl
RAMitchell: It's likely that ternary operators uses predication instead of prediction. We are just speculating without looking at sass code in a profiler, but in my experience this is not an effective optimisation.

RAMitchell:
I see, thanks! Nearest power of four is not ideal from a performance perspective but I think we are still much faster than alternatives and robustness is priority one.

thrust/system/detail/generic/shuffle.inl
djns99: Ah so predication has the latency of both instructions even if no thread in the warp takes the branch?

djns99:
> performance perspective

In theory we could modify the RC5 algorithm slightly and use modulus by a different amount for each register to allow for a power of two instead of 4. However, this would likely require a more comprehensive analysis of the problem to ensure the correct behaviour is maintained, so I dont believe this is the place for this optimisation. There are a range of other candidate ciphers available to consider, but RC5/RC6 (though RC6 rounds to a power of 16) was the main one I found that explicitly supports a variable cipher width.

RAMitchell:
@djns99 let us know when you are ready for final review.

djns99:
If we are ok with RC5 I don't think there is anything left to do at present. So we should be good to go ahead with review and merge.

allisonvacanti:
Whatever implementation you two decide on is fine with me, but it sounds like there are some possible concerns with licensing, too?

Can someone summarize the potentially problematic portions of the current PR, along with their licenses, patents, and other IP concerns?

djns99:
There are two parts that potentially have licensing/IP issues. The first is the Cephes functions used in the test suite, which are directly taken from a NIST test suite for random numbers.
The second potential problem is the round function used for the Feistel Network, of which there are currently two options: RC5 and MurmurHash2

> As for licensing, cephes was taken from the NIST suite for testing randomness, which according to https://csrc.nist.gov/Projects/Random-Bit-Generation/Documentation-and-Software is in the public domain. I took murmur hash from the smhasher repo, the licensing information there say it is also in the public domain.

The Cephes functions are fine as is Murmur Hash if we go that route to the best of my understanding.

> As for licensing, the patent for the algorithm expired in 2015 https://patents.google.com/patent/US5835600A/en so I believe this means we are allowed to use it.

If we go for RC5 (my preferred function @RAMitchell ) then it is covered by an expired patent, with the implementation being my own (based off the original paper https://doi.org/10.1007/3-540-60590-8_7). My understanding is the expired patent means the algorithm is in the public domain, so it is acceptable to use as we have here, however I don't know enough to say for sure.


To the best of my knowledge there are no other licensing/IP issues.

brycelelbach:
I'm going to have one of the NVIDIA open source lawyers look at this.

allisonvacanti:
@djns99 @RAMitchell I haven't been able to get a straight answer on this from our lawyers, and I get the feeling that the only way this patch will move forward is by rewriting it to avoid anything that is known to be associated with a non-NVIDIA owned patent.

Would you be willing to submit a different implementation that avoids the patent-related algorithms?

RAMitchell:
Ok, we will look at rewriting with a different hash. @djns99 if you don't have time, I will take a stab at it later this week.

djns99:
@RAMitchell it would be good if you can. I suspect going back to MurmurHash2 will probably work now I have fixed the tests. Just a bit less academic backing behind it's quality. Should also be able to go to power of two instead of 4 again

allisonvacanti:
Thanks. Sorry about this, but it's the only way to get this moving again in a reasonable amount of time.

allisonvacanti:
Replaced by #1309 due to IP concerns.

PR #1258, opened 2020-08-18
Ensure the CUDA radix sort backend synchronizes before returning

brycelelbach:
Ensure the CUDA radix sort backend synchronizes before returning. Otherwise, copies from temporary storage will race with destruction of said temporary storage.

NVC++ FS #28463

### DIFF
--- a/thrust/system/cuda/detail/sort.h
+++ b/thrust/system/cuda/detail/sort.h
@@ -1598,6 +1598,10 @@ namespace __smart_sort {
     {
       cuda_cub::copy(policy, keys.begin(), keys.end(), keys_first);
     }
+
+    cuda_cub::throw_on_error(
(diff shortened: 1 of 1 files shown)


thrust/system/cuda/detail/sort.h
```diff
@@ -1598,6 +1598,10 @@ namespace __smart_sort {
     {
       cuda_cub::copy(policy, keys.begin(), keys.end(), keys_first);
     }
+
+    cuda_cub::throw_on_error(
+      cuda_cub::synchronize(policy),
+      "merge_sort: failed to synchronize");
```
dkolsen-pgi, line 1604: That should be "smart_sort" or "radix_sort", not "merge_sort".

PR #1264, opened 2020-08-26
Evaluate CUDA_CUB_RET_IF_FAIL macro argument only once

jlowe:
This updates the `CUDA_CUB_RET_IF_FAIL` macro to only evaluate its argument once.  Without this, it can silently suppress CUDA errors in many places in the Thrust code that call it like this: `CUDA_CUB_RET_IF_FAIL(cudaPeekAtLastError());`.

The macro calls `cub::Debug` which in some versions of cub will unconditionally clear the CUDA error.  Therefore when it evaluates `cudaPeekAtLastError()` the second time as the return argument, it will return `cudaSuccess` because the last error was just cleared.  Thus any pending CUDA error before this macro was called with `cudaPeekAtLastErrror()` as the argument will be silently dropped.

### DIFF
--- a/thrust/system/cuda/detail/core/util.h
+++ b/thrust/system/cuda/detail/core/util.h
@@ -652,7 +652,10 @@ namespace core {
   }
 
 #define CUDA_CUB_RET_IF_FAIL(e) \
-  if (cub::Debug((e), __FILE__, __LINE__)) return e;
+  do {                          \
+    auto const error = (e);     \
+    if (cub::Debug(error, __FILE__, __LINE__)) return error; \
+  } while(0);
 
   // uninitialized
   // -------
--- a/thrust/system/cuda/detail/core/util.h
+++ b/thrust/system/cuda/detail/core/util.h
@@ -652,10 +652,10 @@ namespace core {
   }
 
 #define CUDA_CUB_RET_IF_FAIL(e) \
-  do {                          \
+  {                             \
     auto const error = (e);     \
     if (cub::Debug(error, __FILE__, __LINE__)) return error; \
-  } while(0);
+  }
 
   // uninitialized
   // -------


thrust/system/cuda/detail/core/util.h
```diff
@@ -652,7 +652,10 @@ namespace core {
   }
 
 #define CUDA_CUB_RET_IF_FAIL(e) \
-  if (cub::Debug((e), __FILE__, __LINE__)) return e;
+  do {                          \
+    auto const error = (e);     \
+    if (cub::Debug(error, __FILE__, __LINE__)) return error; \
+  } while(0);
```
allisonvacanti, line 659: Nitpick: The `do..while` isn't necessary here, this can just be 

```
{ \
    auto const error = (e);     \
    if (cub::Debug(error, __FILE__, __LINE__)) \
      return error; \
}
```

vuule: @allisonvacanti  There are cases where using a scope block causes issues, please see https://stackoverflow.com/a/1067238

dkolsen-pgi: For that to work, you can't have the semicolon after the `while (0)`.  With that semicolon there, it has the same problem as if it were a block scope.

allisonvacanti: To update this thread with a slack convo, we don't want that behavior here since the old version didn't require a trailing semicolon. Using the `do/while(0)` pattern (without the semicolon) risks breaking other code.

allisonvacanti:
Started CI under DVS CL 28995185.

allisonvacanti:
One DVS builder failed after running out of disk space, but the rest look good.

Just waiting for my local testing script to finish up and then this should be good to go.

allisonvacanti:
All set.

PR #1272, opened 2020-09-08
Fix ambiguous overload when using partition with STL containers

nvibd:
This fixes the error:
```
error: more than one instance of overloaded function "thrust::system::detail::sequential::iter_swap" matches the argument list:
```

When, e.g., using STL containers with `thrust::partition`. Such as in this example: https://www.godbolt.org/z/dfsn6f

This fix has been proposed by @dkolsen-pgi. Many thanks!


allisonvacanti:
LGTM. Testing under internal CL 29046334.

allisonvacanti (approved):

Issue #1273, opened 2020-09-16
[NVBug 3129879] MSVC debug iterators give dreaded "[subobject]" nvcc execution space warnings


allisonvacanti:
## Update:

This is an nvcc bug that will not be fixed. We'll need to disable the treat-warnings-as-errors flags for NVCC + MSVC CUDA Debug targets.

## Original issue:

Encountered while trying to build thrust's `bench.cu` with MSVC 2019:

```
C:\PROGRA~1\NVIDIA~2\CUDA\v11.0\bin\nvcc.exe -forward-unknown-to-host-compiler -DNOMINMAX -DTHRUST_DEVICE_SYSTEM=THRUST_DEVICE_SYSTEM_CUDA -DTHRUST_HOST_SYSTEM=THRUST_HOST_SYSTEM_CPP -I..\ -I..\dependencies\cub -ftemplate-backtrace-limit 64  -gencode arch=compute_75,code=sm_75 -Xcompiler="-Zi -Ob0 -Od /RTC1" -Xcompiler=-MDd -Xcompiler=/W3 -Xcompiler=/WX -Xcompiler=/wd4244 -Xcompiler=/wd4267 -Xcompiler=/wd4800 -Xcompiler=/wd4146 -Xcompiler=/wd4494 -Xcompiler=/bigobj -Xcudafe=--display_error_number -Xcudafe=--promote_warnings -MD -MT internal\benchmark\CMakeFiles\thrust.cpp.cuda.cpp14.bench.dir\bench.cu.obj -MF internal\benchmark\CMakeFiles\thrust.cpp.cuda.cpp14.bench.dir\bench.cu.obj.d -x cu -c ..\internal\benchmark\bench.cu -o internal\benchmark\CMakeFiles\thrust.cpp.cuda.cpp14.bench.dir\bench.cu.obj -Xcompiler=-Fdinternal\benchmark\CMakeFiles\thrust.cpp.cuda.cpp14.bench.dir\,-FS
thrust/detail/allocator/allocator_traits.inl(396): error #3056-D: calling a __host__ function("std::_Iterator_base12::_Iterator_base12") from a __host__ __device__ function("std::_Vector_iterator<    ::std::_Vector_val<    ::std::_Simple_types<char> > > ::_Vector_iterator") is not allowed
```

[...]

PR #1277, opened 2020-09-16
WIP Add bench.cu to CMake build

allisonvacanti:
Currently doesn't compile with MSVC due to #1273.


allisonvacanti:
Closing for now -- I'm experimenting with a simpler benchmark framework here: https://github.com/allisonvacanti/thrust_benchmark

PR #1281, opened 2020-09-18
Update FindTBB.cmake to work with latest MSVC.

allisonvacanti:


brycelelbach (approved):
Why not remove the upper bound?

allisonvacanti:
If `MSVC_VERSION` isn't recognized, it's better to have a clear config-time diagnostic instead of a random link error from using the wrong binaries.

The current failure path doesn't prevent anything from working, there are just some additional `TBB_` CMake variables that need to be set manually when `MSVC_VERSION` isn't recognized.

On a side note, it looks like TBB is getting a proper CMake config soon: https://github.com/oneapi-src/oneTBB/issues/6 Hopefully we'll be able to eventually switch to that and get rid of `FindTBB.cmake` for good.

PR #1286, opened 2020-09-22
Replace test script var THRUST_DIR with THRUST_ROOT.

allisonvacanti:
An internal find_package call sets THRUST_DIR in the CMake cache,
which can cause unexpected behavior during subsequent tests.

Also pass the CMake generator to the test project. This is especially
important on windows.

Update CUB submodule to bring in `add_subdirectory` support as well.


brycelelbach (approved):

allisonvacanti:
Blocked on submodule dep NVIDIA/cub#200.

Issue #1164, opened 2020-05-20
Improve Open-Source Contribution Process


allisonvacanti:
As some of you have noticed, the Thrust and CUB repositories have been going through a lot of changes recently. These have been aimed at improving our open-source contribution process and transitioning our internal development processes to be carried out in the open whenever possible.

A major part of that plan involves migrating our continuous integration testing away from NVIDIA's internal systems and into the open. Additionally, we need to simplify the process of integrating changes from our open-source Github repositories into NVIDIA's internal git and perforce repositories. Automating these processes is complicated for many reasons and will take some time to complete, but they are currently high priority goals for the team.

Currently, it is a rather tedious process to take open-source contributions, test them internally, review them, and integrate them into our public and private repositories. There are a lot of manual steps involved that make it far more difficult than it should be to simply land a PR.

However, once we have open, easy-to-use CI and integration automation in place it will be much easier for us to land your patches. This will allow us to start working through the backlog of pull requests and be more responsive to our open-source contributors. 

As things evolve, we'll keep this issue updated.

allisonvacanti:
# Automatic Testing

We'll want to build/test different configurations for pre-merge testing and post-merge/nightly testing, since the full set of host/device/dialect configurations is very expensive.

## Pre-merge testing

A good set of pre-commit coverage would be cub + all thrust configs with host=`cpp` and dialect=`cpp17`. This can be configured with these multiconfig options:

```
cmake -DTHRUST_INCLUDE_CUB_CMAKE=ON \
      -DTHRUST_ENABLE_MULTICONFIG=ON \
      -DTHRUST_MULTICONFIG_ENABLE_DIALECT_CPP11=OFF \
      -DTHRUST_MULTICONFIG_ENABLE_DIALECT_CPP14=OFF \
      -DTHRUST_MULTICONFIG_ENABLE_DIALECT_CPP17=ON \
      -DTHRUST_MULTICONFIG_ENABLE_SYSTEM_CPP=ON \
      -DTHRUST_MULTICONFIG_ENABLE_SYSTEM_TBB=ON \
      -DTHRUST_MULTICONFIG_ENABLE_SYSTEM_OMP=ON \
      -DTHRUST_MULTICONFIG_ENABLE_SYSTEM_CUDA=ON \
      -DTHRUST_MULTICONFIG_WORKLOAD=SMALL \
      /path/to/thrust
```

[...]

PR #1296, opened 2020-09-26
Add gpuCI CPU-only script, based on RMM's gpuCI scripts

brycelelbach:


ci/build_only/build.bash
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env bash
```
raydouglass, line 1: This looks great! My only comment is that by gpuCI convention we name this file `ci/cpu/build.sh` if that's okay with you.

brycelelbach: I wasn't quite sure what to do here; we do have some non-CUDA backends that do need to run tests on CPUs. I guess I was imaging we would have:

- `ci/build_only/build.bash`, which would build everything, and down the road would export the test binaries to run on GPU nodes.
- `ci/cpu/build.bash`, which would just build CPU backends and would run their tests.
- `ci/gpu/build.bash`, which would just build GPU backends and run their tests.

But maybe I'm going about this wrong? What do you think?

brycelelbach: Perhaps we should have just the two configurations, and have them be fully parameterized? @allisonvacanti, could we perhaps set up our CMake configuration so that CTest only runs GPU tests if the node has GPUs supported? We wouldn't need to detect this - we could have a CMake variable 'THRUST_RUN_CUDA_TESTS', or something like that.

Then we would have:

- `ci/cpu/build.bash`: Builds a parameterized set of configurations, then runs all tests (expect CUDA tests), and packages up the tests for redistribution.
- `ci/gpu/build.bash`: Builds a parameterized set of configurations, then runs all tests.

allisonvacanti: > could we perhaps set up our CMake configuration so that CTest only runs GPU tests if the node has GPUs supported

No need to modify the CMake config for this -- we can do that today with a -R [regex] option to ctest. All GPU tests can be run with `ctest -R "^cub|^thrust.*cuda"`.

Similarly, all CPU tests can be run with the -E [exclude regex] option `ctest -E "^cub|^thrust.*cuda"`.

If we wanted to simplify this we could explicitly tag the tests with `GPU` and `CPU` labels and run them with `ctest -L GPU`, etc, too.

Building today can also be targeted by listing out some per-`host.device.dialect` meta targets in the scripts:

```
# Build all cuda targets:
ninja thrust.cpp.cuda.cpp11.all \
      thrust.cpp.cuda.cpp14.all \
      thrust.cpp.cuda.cpp17.all \
      thrust.omp.cuda.cpp11.all \
      [ ... ] \
      thrust.tbb.cuda.cpp11.all \
      [ ... ] \
      cub.all
```

Similar to the ctest labels, it'd be trivial to add `thrust.gpu.all`, `thrust.cpu.all` metatargets that would simplify this if needed.

allisonvacanti:
I jotted down some thoughts about what we should build for pre/post merge testing here: https://github.com/NVIDIA/thrust/issues/1164#issuecomment-703163439

brycelelbach:
@allisonvacanti should we have this script only build/test Thrust, and then have a separate script that builds/tests CUB, in the CUB repo?

Yet another argument for mono-repo.

ci/build_only/build.bash
```diff
@@ -0,0 +1,81 @@
...
+if [[ -z "${CUDACXX}" ]] ; then
+  export CUDACXX="nvcc"
+fi
+
+CMAKE_FLAGS="-DCMAKE_CXX_COMPILER=${CXX} -DCMAKE_CUDA_COMPILER=${CUDACXX}"
+
+# If it's a nightly build, build all configurations.
+if [[ "${BUILD_MODE}" = "branch" ]] ; then
+  CMAKE_FLAGS="${CMAKE_FLAGS} -DTHRUST_MULTICONFIG_WORKLOAD=FULL"
+fi
```
brycelelbach, line 48: We should get rid of this and just use "c++" and "cu++", which the container will set up.

allisonvacanti:
If at all possible, I'd like to just have a script in thrust to reduce the maintenance burden. Thrust can build and run the cub tests by setting THRUST_INCLUDE_CUB_CMAKE (or something similar). I'd want cub CI to build the thrust CUDA tests, too, so just having one script somewhere that both projects use would be ideal imo.

Monorepo would def make this easier 🙂

brycelelbach:
All hail our future monorepo overlord.

allisonvacanti:
🙌🙌🙌

Issue #1305, opened 2020-10-07
Binary search can fail due to overflow error


r-barnes:
The binary search implementation suffers from a [well-known issue](https://en.wikipedia.org/wiki/Binary_search_algorithm#Implementation_issues) that can lead to incorrect results on large datasets. This happens [here](https://github.com/NVIDIA/thrust/blob/main/thrust/system/detail/generic/scalar/binary_search.inl#L55) and [here](https://github.com/NVIDIA/thrust/blob/main/thrust/system/detail/generic/scalar/binary_search.inl#L97). There's a simple fix which I'll provide a PR for.

r-barnes:
@allisonvacanti, #1306 has been approved for merge, which resolves this bug. Not sure how that should be handled on your end.

r-barnes:
Oh, nevermind, I see that you're on that thread too.

PR #1306, opened 2020-10-07
Fix binary search middle calculation to avoid overflows

r-barnes:
Addresses #1305


griwes (approved):
A really nice fix for a somewhat embarrassing bug, thanks ;)

allisonvacanti:
DVS CL 29189207.

PR #1310, opened 2020-10-12
Variadic tuple preparation

andrewcorrigan:
Some simplifications preparing Thrust for a variadic tuple implementation (some day... #524).   Other changes would require a bit more coordination and can come separately, assuming these sorts of changes are now mergeable.

With ` -DTHRUST_DEVICE_SYSTEM=CPP ` I get:
```
100% tests passed, 0 tests failed out of 151

Total Test time (real) = 108.10 sec
```

Issue #1212, opened 2020-06-30
Switch to `cuda::std::tuple`/`<cuda/std/tuple>`


brycelelbach:
Note that this requires us to sort out fetching the libcu++ headers in Thrust/CUB internal builds.

andrewcorrigan:
I'm hoping to try this out.  Is there a branch anywhere implementing any of this?  As a warm-up for `cuda::std::tuple`, I tried swapping in `std::tuple` and `std::pair`:

In [thrust/pair.h](../tree/main/thrust/pair.h):
```
namespace thrust
{
using std::pair;
using std::make_pair;
}
```

In [thrust/tuple.h](../tree/main/thrust/tuple.h):
```
namespace thrust
{
using std::tuple;
using std::make_tuple;
using std::get;
using std::tuple_size;
using std::tuple_size_v;
using std::tuple_element;
using std::tuple_element_t;
using std::tie;
using std::tuple_cat;
using std::index_sequence;
using std::make_index_sequence;
}
```

However, I keep getting stuck on specializations related to `detail::tuple_of_iterator_references`, like the following:
```
/thrust/thrust/iterator/detail/tuple_of_iterator_references.h:124:8: error: class template partial specialization of 'tuple_size' not in a namespace enclosing '__1'
struct tuple_size<detail::tuple_of_iterator_references<Ts...>>
       ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/__tuple:24:50: note: explicitly specialized declaration is here
template <class _Tp> struct _LIBCPP_TEMPLATE_VIS tuple_size;
                                                 ^
```

I'm not a C++ expert and am not sure what the right way to specialize things from `std` is.  Therefore, I'm wondering: is there a branch already handling all this for `cuda::std::tuple`? Alternatively, if either of you @brycelelbach or @allisonvacanti might have any advice, I would appreciate it greatly.

brycelelbach:
So this should be doable now that libcu++ can be used from Thrust.

I knew the tuple of references thing was going to be a challenge. I'll defer to @allisonvacanti and @griwes to sort out what we should do here.

andrewcorrigan:
It's exciting to hear this should now be doable.  I hope I can be of assistance making it happen.  Since I posted a few hours ago, I tried moving the specialization themselves into `std`:

```
namespace std
{
// define tuple_size, tuple_element, etc.
template<class... Ts>
[...]

PR #1311, opened 2020-10-12
Preparing zip_iterator and tuple_of_iterator_references for variadic tuple

andrewcorrigan:
Depends on #1310.  This reduces dependence of `zip_iterator` on implementation details of `thrust::tuple`, such as `head_type`, `tail_type`, and `null_type`.

The reason I submitted this as a separate PR is the following.  Unfortunately, certain implementation details remain very inconvenient to workaround. Therefore, I introduced a guard `THRUST_VARIADIC_TUPLE`, which is obviously disabled.  Upon a future variadic `thrust::tuple` (regardless of whether it's from [libcudacxx](https://github.com/NVIDIA/libcudacxx), `std::tuple`, [jaredhoberock/tuple](https://github.com/jaredhoberock/tuple), etc.), the code guarded by `THRUST_VARIADIC_TUPLE` can be activated, while the `#else` branch code can be removed.  The code guarded by `THRUST_VARIADIC_TUPLE` works.  I have tested it extensively over the past 5 years on my [variadic branch](https://github.com/andrewcorrigan/thrust/tree/variadic).

With `-DTHRUST_DEVICE_SYSTEM=CPP`:
```
100% tests passed, 0 tests failed out of 151

Total Test time (real) = 109.96 sec
```


allisonvacanti:
The split between these PRs makes sense to me. I'll focus on getting the other one landed, and probably hold off on adding this until we have the new tuple in-place and can remove the preprocessor guards.

allisonvacanti:
For bookkeeping, this PR is blocked on #1212.

andrewcorrigan:
In that case, since it's blocked anyway, I'll redo the commit without the guard.  Then, once #1212 is taken care of, it can be merged as-is.   I also have code ready to go for `tuple_of_iterator_references`, which is directly related since it used by `zip_iterator` and similarly blocked (related to incompatibility between non-variadic and variadic `tuple` constructors). I would like to build that change also into this PR.

FYI, in case you're interested in seeing these changes in action, I created [a new variadic tuple branch](https://github.com/andrewcorrigan/thrust/tree/variadic-2020).  I tried building and running all the tests.  They succeeded with one exception, that I disabled.

andrewcorrigan:
The last commit fixes the one test that I mentioned had failed.  All tests now pass using these changes on [my branch](https://github.com/andrewcorrigan/thrust/tree/variadic-2020).

andrewcorrigan:
@allisonvacanti  It looks like the reason for the incompatibility was #1314.  I was not expecting such a simple fix. With #1314 merged, I think that this PR could be ready to go as soon as #1310 is merged.

Would it make more sense to add these commits to #1310 and close this PR?  Or should I keep it separate?

allisonvacanti:
Ahh, good catch, that makes some amount of sense.

It'd probably be simpler to just combine the PRs and remove the workaround. We have some time -- our internal CI is currently unusable so I won't be able to move this forward until sometime next week at the earliest anyway.

PR #1317, opened 2020-10-14
gpuCI: Local docker builds and gpuCI GPU support

brycelelbach:
- Add support for local docker builds in a gpuCI-like environment.
- Rename the existing `ci/cpu/build.bash` to `ci/common/build.bash`.
- Create two new scripts, `ci/{cpu,gpu}/build.bash`, that source `ci/common/build.bash`.
- Fix uses of the wrong Bash variable name.


ci/common/build.bash
```diff
@@ -0,0 +1,89 @@
...
+# Switch to project root; also root of repo checkout.
+cd ${WORKSPACE}
+
+# If it's a nightly build, append current YYMMDD to version.
+if [ "${BUILD_MODE}" == "branch" ]; then
+  export VERSION_SUFFIX=`date +%y%m%d`
+fi
+
+# The Docker image sets up `${CXX}` and `${CUDACXX}`.
+CMAKE_FLAGS="-DCMAKE_CXX_COMPILER=${CXX} -DCMAKE_CUDA_COMPILER=${CUDACXX}"
```
allisonvacanti, line 41: Should have quotes around each option here -- these will break if there's a space in the path to the compiler.

ci/common/build.bash
```diff
@@ -0,0 +1,89 @@
...
+
+# If it's a nightly build, build all configurations.
+if [ "${BUILD_MODE}" == "branch" ]; then
+  CMAKE_FLAGS="${CMAKE_FLAGS} -DTHRUST_MULTICONFIG_WORKLOAD=FULL"
```
allisonvacanti, line 45: `THRUST_MULTICONFIG_WORKLOAD=FULL` has no effect unless multiconfig builds are enabled. We should also explicitly specify the dialects and systems we want to test and tell CMake to include CUB's tests.

I've documented Thrust's CMake options [here](https://github.com/NVIDIA/thrust/blob/main/CONTRIBUTING.md#cmake-options). See [this comment](https://github.com/NVIDIA/thrust/issues/1164#issuecomment-703163439) for an example multiconfig setup.

ci/local/build.bash
```diff
@@ -0,0 +1,183 @@
...
+
+################################################################################
+# Thrust and CUB local containerized build script
+################################################################################
+
+function usage {
+  echo "Usage: ${0} [flags...] [cmake-targets...]"
+  echo
+  echo "Build and test your local repository using a gpuCI Docker image."
+  echo "If CMake targets are specified, only those targets are built and tested."
```
allisonvacanti, line 16: FYI, this will only work with non-meta targets, since CTest doesn't know about meta targets.

This means that targets for specific tests (`thrust.cpp.cuda.cpp17.test.scan`) will work, but entire configs (`thrust.cpp.cuda.cpp17.all`) will not. If you want to latter to work let me know and we can figure something out.

ci/common/build.bash
allisonvacanti: Also -- this may be why you were seeing the warnings about

```
You have changed variables that require your cache to be deleted.
Configure will be re-run and you may have to reset some variables.
```

If there is an existing CMakeCache.txt in the build directory that was configured for a different compiler, CMake will get confused, give up, and start over with a clean configuration. Most options can be changed when re-invoking CMake on an existing cache, but some of the compiler specific ones are problematic.

allisonvacanti (changes requested):
The multiconfig stuff needs to be fixed up for sure, the rest are just FYIs/suggestions.

I'm looking forward to trying this out!

ci/common/build.bash
```diff
@@ -0,0 +1,88 @@
...
+fi
+
+CTEST_FLAGS=""
+
+if [ "${BUILD_KIND}" == "cpu" ]; then
```
raydouglass, line 50: There is a gpuCI environment variable for this called `BUILD_TYPE`. Check out all of the available variables here: https://docs.rapids.ai/maintainers/gpuci/#environment-variables

ci/common/determine_build_parallelism.bash
```diff
@@ -0,0 +1,35 @@
+#! /usr/bin/env bash
```
raydouglass, line 1: This is very cool

ci/common/build.bash
brycelelbach: Quotes have been added.

ci/common/build.bash
brycelelbach: Ah cool thanks.

ci/common/determine_build_parallelism.bash
brycelelbach: :) @allisonvacanti wrote most of it.

ci/local/build.bash
brycelelbach: I'd like the latter to work. Should I file an issue?

allisonvacanti: Please do.

To get partway there, I think we can just drop the `.all` from the metatargets. Then `ninja thrust.cpp.cuda.cpp17` and `ctest -R thrust.cpp.cuda.cpp17` would both work as expected.

There's another issue -- doing multiple configs at once will be tricker because `ctest -R` behaves kinda weird. We'd have to join the configs into a regex to get it to actually run all tests from those configs:

```bash
# Works easily on ninja:
ninja thrust.cpp.cuda.cpp11 thrust.cpp.cuda.cpp14 thrust.cpp.cuda.cpp17
# Doesn't work, -R only accepts a single argument:
ctest -R thrust.cpp.cuda.cpp11 thrust.cpp.cuda.cpp14 thrust.cpp.cuda.cpp17
# Doesn't work, only one -R can be specified, the final overwrites the others and only cpp17 tests run
ctest -R thrust.cpp.cuda.cpp11 -R thrust.cpp.cuda.cpp14 -R thrust.cpp.cuda.cpp17
# Works, but it's kinda horrible because it's really a regex:
ctest -R "thrust.cpp.cuda.cpp11|thrust.cpp.cuda.cpp14|thrust.cpp.cuda.cpp17"
# What we are really trying to say:
ctest -R "^thrust\.cpp\.cuda\.cpp11|^thrust\.cpp\.cuda\.cpp14|^thrust\.cpp\.cuda\.cpp17"
```

So some bash string manip will be needed to get to something useable for multiple targets.

allisonvacanti: Actually, it looks like multiple -R regexes can be combined when -U is passed. So this will also work (though again, being sloppy with regexes):

```
ctest -U -R thrust.cpp.cuda.cpp11 -R thrust.cpp.cuda.cpp14 -R thrust.cpp.cuda.cpp17
```

TIL the "-U" is for union, and there's also a "-I" for intersection.

Not sure if this buys us anything since we'd still have to process the string, but it's definitely cleaner than generating a single mega-regex.

brycelelbach: I've fixed this for now, it will work with either no target or specific targets.

ci/common/build.bash
```diff
@@ -0,0 +1,116 @@
...
+fi
+
+CTEST_FLAGS=""
+
+if [ "${BUILD_TYPE}" == "cpu" ]; then
+  CTEST_FLAGS="${CTEST_FLAGS} -E ^cub|^thrust.*cuda"
+fi
+
+if [ ! -z "${@}" ]; then
+  CTEST_FLAGS="${CTEST_FLAGS} -R ^${@}$"
```
allisonvacanti, line 86: Two things to be aware of:

1. This only supports a single target.
2. This won't work with configuration metatargets because of the trailing `$`

allisonvacanti (approved):

Issue #1294, opened 2020-09-25
CUDA Toolkit packaging does not include Thrust's CMake find_package config files


allisonvacanti:
We also need to make sure that these are located in a proper location once they're added (see #1292).

allisonvacanti:
Adding a reminder for myself to talk with Barry Fujii about this once the CTK is using a version of Thrust with #1323 applied.

allisonvacanti:
The CMake package will be included with an upcoming CTK release.

PR #1323, opened 2020-10-16
Update install rules to install cmake into lib directory

### LINKED ISSUE #1292: thrust-config.cmake is installed in a non-standard location
Usually when configuring a project that provides a cmake package config file with `cmake -DCMAKE_INSTALL_PREFIX=$PFX`, I can later pull that into a downstream project using `find_package(proj)`. In order for it to be found, I set `CMAKE_PREFIX_PATH=$PFX` (well, mostly my package manager does it, but anyway).

Thrust nicely provides a `thrust-config.cmake` now that's installed together with the headers, but the file itself is put into a non-standard location, ie., `$PFX/include/cmake/thrust/thrust-config.cmake`.  `cmake` searches in various locations (see https://cmake.org/cmake/help/latest/command/find_package.html),  but not there. Most commonly I see `$PFX/lib/cmake/pkg/pkg-config.cmake` being used, though since thrust is header-only, maybe `share` instead of `lib` would be more appropriate.

Somewhat related to do this, the Thrust README file recommends using this to help cmake find Thrust:
```
$ cmake . -DThrust_DIR=/usr/local/cuda/include/thrust/cmake/
```

I believe the newer, more standard way of providing paths for a specific package (besides `CMAKE_PREFIX_PATH`) is
```
$ cmake . -DThrust_ROOT=/usr/local/cuda
```
(which again wouldn't work because of the non-standard location of the config file in the include directory)

### LINKED ISSUE #1322: CMake files are installed in non-searched location
Currently the cmake files are installed in `include/thrust/cmake` which isn't a standard search location: https://cmake.org/cmake/help/v3.18/command/find_package.html#search-procedure

I think projects typically install in `lib/${package_name}/cmake` even if they're header only. I.E. `spdlog` does this today.

kkraus14:
Fixes #1292

Updates the cmake install rules to install in `lib/thrust/cmake` instead of `include/thrust/cmake` as the include folder is not searched in a normal `find_package(Thrust)` call.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,18 @@ else()
   set(THRUST_TOPLEVEL_PROJECT OFF)
 endif()
 
+# This must be done before any languages are enabled:
+if (THRUST_TOPLEVEL_PROJECT)
+  include(cmake/ThrustCompilerHacks.cmake)
+endif()
+
+# This must appear after our Compiler Hacks or else CMake will delete the cache
+# and reconfigure from scratch.
+# This must also appear before the installation rules, as it is required by the
+# GNUInstallDirs CMake module.
+enable_language(CXX)
+
+# Optionally include installation rules for non-top-level builds:
 option(THRUST_ENABLE_INSTALL_RULES "Enable installation of Thrust" ${THRUST_TOPLEVEL_PROJECT})
 if (THRUST_ENABLE_INSTALL_RULES)
   include(cmake/ThrustInstallRules.cmake)
@@ -30,13 +42,6 @@ if (NOT THRUST_TOPLEVEL_PROJECT)
   include(cmake/ThrustAddSubdir.cmake)
   return()
 endif()
-
-include(cmake/AppendOptionIfAvailable.cmake)
-
-include(cmake/ThrustBuildCompilerTargets.cmake)
-include(cmake/ThrustBuildTargetList.cmake)
-include(cmake/ThrustMultiConfig.cmake)
-include(cmake/ThrustUtilities.cmake)
 
 option(THRUST_ENABLE_HEADER_TESTING "Test that all public headers compile." "ON")
 option(THRUST_ENABLE_TESTING "Build Thrust testing suite." "ON")
@@ -52,6 +57,12 @@ if (NOT (THRUST_ENABLE_HEADER_TESTING OR
          THRUST_INCLUDE_CUB_CMAKE))
   return()
 endif()
+
+include(cmake/AppendOptionIfAvailable.cmake)
+include(cmake/ThrustBuildCompilerTargets.cmake)
+include(cmake/ThrustBuildTargetList.cmake)
+include(cmake/ThrustMultiConfig.cmake)
+include(cmake/ThrustUtilities.cmake)
 
 # Add cache string options for CMAKE_BUILD_TYPE and default to RelWithDebInfo.
 if ("" STREQUAL "${CMAKE_BUILD_TYPE}")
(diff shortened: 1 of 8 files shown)


germasch:
I was going to suggest something like this, so I like it ;) 

I think, however, there may be an issue with how thrust is packaged, including, I suspect, internally at Nvidia. That is, the directory structure of this repo is such that one can just copy `thrust/` to `somewhere/include/thrust/` and `cmake` or `make install` are never even used. It looks like this required some non-standard handling of things, including have the `.cmake` files under `include/thrust/` and also, e.g., not having a standard generated version file, but rather one that parses the version from the thrust header (which is elegant, but different from what's usually done).

I do strongly support having the installed cmake files in a standard location where they'll be found. I just can't see a way of achieving this without getting inconsistencies between the installed and uninstalled version of thrust. Maybe a somewhat less confusing way would be to move the `.cmake` files out of `thrust/cmake/`, so in that case at least they wouldn't appear in two different places depending on how thrust is used, but only the installed version would have them, under `lib/cmake/`. But obviously, in that case it's important to make sure that the CUDA toolkit and whatever SDKs ship with these config files, not just whatever is under `include/thrust`.

allisonvacanti:
Our internal CTK packaging scripts don't pick these files up at all at the moment (#1294), so this won't break anything there.

I agree that this is the way to go -- leave them where they are in the source tree to have some measure of support for the legacy "just copy the `thrust` dir" usecase, but put them somewhere discoverable for installations.

I'll give this a try sometime this week and hopefully get it into 1.11.0 if all goes well.

allisonvacanti:
@kkraus14 @germasch I've updated this (and NVIDIA/cub#217) to work for both source tree and install tree, and added a test to validate the install tree package.

If you get a chance to try this out, please let me know if you find any issues.

germasch:
@allisonvacanti: This PR works for me, thanks!

I kinda like it better to have the config files in `lib/cmake/{thrust,cub}` rather than `lib/{thrust,cub}/cmake`, probably because I usually see the former. But either way works. Like you, I guess, I don't like all the empty directories. This is a way to avoid that: (And it also reverses to `lib/cmake`, though that's not the main point.)

```diff
diff --git a/cmake/ThrustInstallRules.cmake b/cmake/ThrustInstallRules.cmake
index 31db9142..a69ec78d 100644
--- a/cmake/ThrustInstallRules.cmake
+++ b/cmake/ThrustInstallRules.cmake
@@ -8,13 +8,8 @@ install(DIRECTORY "${Thrust_SOURCE_DIR}/thrust"
     PATTERN "*.inl"
 )

-# Unfortunately this creates a ton of empty directories, see issue:
-# https://gitlab.kitware.com/cmake/cmake/-/issues/19189
-install(DIRECTORY "${Thrust_SOURCE_DIR}/thrust"
-  TYPE LIB
-  FILES_MATCHING
-    PATTERN "cmake/*.cmake"
-    PATTERN "cmake/*.md"
+install(DIRECTORY "${Thrust_SOURCE_DIR}/thrust/cmake/"
+  DESTINATION "lib/cmake/thrust"
 )

 # Depending on how Thrust is configured, CUB's CMake scripts may or may not be
@@ -29,11 +24,7 @@ if (THRUST_INSTALL_CUB_HEADERS)
       PATTERN "*.cuh"
   )

-  # Unfortunately this creates a ton of empty directories, see issue:
-  # https://gitlab.kitware.com/cmake/cmake/-/issues/19189
-  install(DIRECTORY "${Thrust_SOURCE_DIR}/dependencies/cub/cub"
-    TYPE LIB
-    FILES_MATCHING
-      PATTERN "cmake/*.cmake"
+  install(DIRECTORY "${Thrust_SOURCE_DIR}/dependencies/cub/cub/cmake/"
+    DESTINATION "lib/cmake/cub"
   )
 endif()
```

allisonvacanti:
That sounds fine to me. I avoided that the first time around because it was clunky to make work with the `TYPE LIB` specification, which handles packaging usecases where the desired library path isn't `[prefix]/lib`.

It looks like CMake also provides the target library path as part of the [`GNUInstallDirs`](https://cmake.org/cmake/help/v3.15/module/GNUInstallDirs.html#module:GNUInstallDirs) module.

I'll make the suggested change with the addition of the `${CMAKE_INSTALL_LIBDIR}` variable.

I agree that switching the `thrust/cmake` nesting is cleaner, especially if we aren't bound by the `TYPE LIB` pattern.

germasch:
That sounds great. I was thinking to use GNUInstallDirs to not hardcode the `lib`, but wasn't sure that it's worth the additional (slight) complication, but it is the right way of going about this.

allisonvacanti:
It took a bit of refactoring to meet the `GNUInstallDirs` requirements, but this last push fixes up the install rules as proposed by @germasch. I'll merge this sometime this week before the 1.11.0 release.

allisonvacanti:
Last push just adjusts the new submodule SHA. All set!

allisonvacanti (approved):

PR #1329, opened 2020-10-24
cross_system_copy_n: Dont attempt a copy if we're tyring to copy 0 elements

### LINKED ISSUE #1275: CPU->GPU copy of a default-initialized vector [windows/vs2019]
In Windows, with VS2019 16.7.2 (debug mode) and the default thrust provided with CUDA 11 package, I found this problem:

```#include <thrust/device_vector.h>
int main()
{
    thrust::device_vector<int> gpu;
    std::vector<int> cpu;
    gpu = cpu; // assertion in debug
}
```

The code triggers an assertion in `vector:46` (standard vector header, provided by Microsoft) coming from `thrust/system/cuda/detail/internal/copy_cross_system.h:106` (it's inside the function `cross_system_copy_n`, and the assert is triggering because the code is dereferencing the `begin()` iterator, which is invalid (being the vector empty), and more specifically is nullptr (being uninitialized).

The problem disappears if the `gpu` vector is allocated (e.g. by pushing back and clearing) before the assignment:

```

int main()
{
    thrust::device_vector<int> gpu;
    std::vector<int> cpu;
    gpu.push_back(1);
    gpu.clear();
    gpu = cpu; // works! (actually, the CUDA11 version triggers a different assertion, but that has been fixed already)
}
```

I think this might be loosely related to #938 and #939 .

The other assertion I mentioned above has been fixed by @allisonvacanti here (https://github.com/thrust/thrust/commit/b9a20738fd68ce4af4dd2df7b1820bf9a2fc69dc)

bjude:
Attempting to perform this copy with 0 elements caused a debug assertion when compiling with MSVC in debug mode.

Fixes #1275

### DIFF
--- a/testing/vector.cu
+++ b/testing/vector.cu
@@ -645,6 +645,19 @@ DECLARE_VECTOR_UNITTEST(TestVectorReserving)
 
 
 template <class Vector>
+void TestVectorUninitialisedCopy(void)
+{
+    thrust::device_vector<int> v;
+    std::vector<int> std_vector;
+
+    v = std_vector;
+
+    ASSERT_EQUAL(v.size(), static_cast<size_t>(0));
+}
+DECLARE_VECTOR_UNITTEST(TestVectorUninitialisedCopy);
+
+
+template <class Vector>
 void TestVectorShrinkToFit(void)
 {
     typedef typename Vector::value_type T;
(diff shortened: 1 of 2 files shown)


allisonvacanti:
LGTM -- thanks for the patch!

allisonvacanti:
Last push fixed a signed/unsigned comparison warning from gcc 7.

brycelelbach:
ok to test

GPUtester:

brycelelbach:
run tests

allisonvacanti (approved):

PR #1346, opened 2020-11-23
Enable more host / c++ compilers for gpuCI builds.

allisonvacanti:
Adds all supported major version of GCC and Clang.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,16 @@
+# Exclude these commits from git-blame and similar tools.
+#
+# To use this file, run the following command from the repo root:
+#
+# ```
+# $ git config blame.ignoreRevsFile .git-blame-ignore-revs
+# ```
+#
+# Include a brief comment with each commit added, for example:
+#
+# ```
+# d92d9f8baac5ec48a8f8718dd69f415a45efe372 # Initial clang-format
+# ```
+#
+# Only add commits that are pure formatting changes (e.g.
+# clang-format version changes, etc).
--- a/ci/axis/cpu.yml
+++ b/ci/axis/cpu.yml
@@ -14,5 +14,15 @@ CXX_VER:
   - 10
 
 exclude:
+  # Unsupported compiler version
   - CXX_TYPE: clang
     CXX_VER: 5
+  # This config is broken in the docker image: https://github.com/NVIDIA/cccl/issues/6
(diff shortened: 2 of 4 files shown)


allisonvacanti:
@raydouglass @brycelelbach Heads up that I'll be testing out Thrust/CUB with the complete set of supported GCC and Clang compilers this week. This adds 10 additional builds.

run tests

allisonvacanti:
@raydouglass @brycelelbach There are some config related issues that I'll your help with:

1. Clang 6 and GCC 10 are both failing because the `CXX` bash var is not set before our `build.bash` executes. I can't find where this is supposed to be set, can either of you fix this (or point me at what needs fixing)?
2. Clang 10 fails because the installed `nvcc` doesn't support that version of clang.

raydouglass:
> 1. Clang 6 and GCC 10 are both failing because the `CXX` bash var is not set before our `build.bash` executes. I can't find where this is supposed to be set, can either of you fix this (or point me at what needs fixing)?

https://github.com/NVIDIA/cccl/blob/main/docker/Dockerfile#L133

It looks like `/etc/cccl.bashrc` has empty `CC` and `CXX` variables:
```
$ docker run -it --rm gpuci/cccl:cuda11.0-devel-ubuntu18.04-clang6 cat /etc/cccl.bashrc
export CC=
export CXX=
export CUDACXX=/usr/local/cuda/bin/nvcc
```

Perhaps clang6 and gcc10 install themselves as `clang` or `gcc` instead of a versioned binary?

allisonvacanti:
> Perhaps clang6 and gcc10 install themselves as clang or gcc instead of a versioned binary?

I checked my Ubuntu 18.04 box, and `clang10` is just doing it's own thing with naming conventions:

```
clang++-6.0
clang++-7
clang++-8
clang++-9
clang++-10
```

So that's probably it.

`g++-10` does exist, though I'm pulling that one from the test toolchain PPA: https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test Maybe we need to add that to the docker image as well.

allisonvacanti:
run tests

allisonvacanti:
run tests

allisonvacanti:
run tests

brycelelbach (approved):

allisonvacanti:
DVS CL: 29371612

PR #1350, opened 2020-12-04
Add gpuCI GPU testing, automatic GPU detection, and support for more compilers

brycelelbach:
Add gpuCI GPU testing, automatic GPU detection, and support for more compilers:

* CMake: Add support for detecting the compute archs of the GPUs in your system
  at configure time.
* gpuCI: Add a GPU node configuration that builds and tests as little as possible.
* gpuCI: Cleanup logic for different build and test configurations.
* gpuCI: Fix an unfortunate typo in `determine_build_parallelism.bash` which
  led to the parallelism level not being set.
* gpuCI: Add support for NVC++.
* gpuCI: Update to CUDA 11.1 and Ubuntu 20.04.
* gpuCI: Add NVC++ and ICC configurations to the CPU axis file.
* gpuCI: Add a GPU axis file.
* gpuCI: Increase the desired memory per build thread to 4GB.
* gpuCI: Add a -j switch which controls build parallelism to `ci/local/build.bash`.

PR #1351, opened 2020-12-05
Enable gpuCI builds for icc / nvc++.

allisonvacanti:
gcc-10 has also been re-enabled since NVIDIA/cccl#8 is fixed.

### DIFF
--- a/ci/axis/cpu.yml
+++ b/ci/axis/cpu.yml
@@ -1,9 +1,19 @@
 OS_VER:
   - ubuntu18.04
 
+SDK_TYPE:
+  - nvidia/cuda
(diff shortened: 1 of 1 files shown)


allisonvacanti:
run tests

brycelelbach:
The nvcxx build isn't triggering because the image is getting built with the wrong name, we're looking into that. It should be called `nvhpc20.7-devel-centos7-nvcxx20.7` but instead it's called `cuda20.7-devel-centos7-nvcxx20.7`.

As for the icc container, I'm not sure, I think we'll have to ask @raydouglass.

brycelelbach:
Superseded by #1350

PR #1358, opened 2020-12-10
CMake: Remove old references to the "Feta" codename for NVC++.

brycelelbach:


brycelelbach:
ok to test

CMakeLists.txt
```diff
@@ -43,7 +43,7 @@ if (NOT THRUST_TOPLEVEL_PROJECT)
   return()
 endif()
 
-option(THRUST_ENABLE_HEADER_TESTING "Test that all public headers compile." "ON")
+option(THRUST_ENABLE_HEADER_TESTING "Test that all public headers compile." ON)
```
allisonvacanti, line 46: Both forms are valid, but why change one and not the others?

allisonvacanti (approved):

CMakeLists.txt
brycelelbach: I don't think I did this intentionally, it was just something I copied pasted.

brycelelbach: Fixed.

PR #1362, opened 2020-12-15
Replace the term "sanity", which is non-inclusive.

brycelelbach:


examples/CMakeLists.txt
```diff
@@ -28,8 +28,8 @@ if (THRUST_ENABLE_EXAMPLE_FILECHECK)
   endif()
 
   execute_process(
-    COMMAND "${THRUST_FILECHECK_EXECUTABLE}" "${filecheck_data_path}/thrust.sanity.filecheck"
```
allisonvacanti, line 31: Did this configure successfully when you tested it? I believe that this file in `internal/test/` will need to be updated too or this check will start failing. `git grep` shows a few other usages in that directory, too.

tomilov:
What does it mean "non-inclusive"? Which property is not included into the meaning namely?

examples/CMakeLists.txt
brycelelbach: Fixed.

brycelelbach:
> What does it mean "non-inclusive"? Which property is not included into the meaning namely?

The term "sanity", which has denotative (e.g. primary) meaning of "basic tests", also has negative connotations (e.g. secondary meanings) relating to mental health. The use of the term "sanity check" may suggest that individuals with mental illnesses are inferior, wrong, or incorrect.

In accordance with NVIDIA's code of conduct, we should strive to avoid non-inclusive terms like "sanity" and use neutral and inclusive language instead.

The term "sanity" is listed as a non-inclusive term to avoid by a number of technical and standards organizations:

* W3C: https://w3c.github.io/manual-of-style/#inclusive
* IETF: https://github.com/ietf/terminology
* Google Engineering: https://developers.google.com/style/inclusive-documentation
* Twitter Engineering
* INCITS (US national standards)
* ISO (international standards)

brycelelbach:
run tests

PR #1373, opened 2021-01-07
Fix compilation error when a standard library type is wrapped in `thrust::optional`

vuule:
Declaring a `thrust::optional<std::string>` causes a following error:
`call of overloaded ‘addressof(std::__cxx11::basic_string<char>&)’ is ambiguous`

This is because both `thrust` and `std` have `addressof` (`std::addressof` is found through ADL).

The patch is to simply explicitly use `thrust::addressof` in this context. Verified locally to fix the issue.

### DIFF
--- a/thrust/optional.h
+++ b/thrust/optional.h
@@ -491,7 +491,7 @@ template <class T> struct optional_operations_base : optional_storage_base<T> {
   template <class... Args>
   __host__ __device__
   void construct(Args &&... args) noexcept {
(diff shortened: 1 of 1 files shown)


allisonvacanti:
Looks good to me, thanks for the patch!

allisonvacanti:
run tests

allisonvacanti:
DVS CL: 29489327

allisonvacanti (approved):

PR #1381, opened 2021-02-06
Skip build/test step on nvc++ gpuci until a future release.

allisonvacanti:
nvc++ 20.9 builds time-out on gpuci, and the build fails when repro'd locally due to known issues in the HPC SDK. Until we have a public version of the SDK that can build thrust/cub cleanly without timing out, skip the build/test steps.

Usually when we break nvc++, we break the cmake configuration, so I've left that step enabled so this CI config still has some value.


allisonvacanti:
run tests

allisonvacanti:
Blocked; see if the CI update fixes it first.

PR #1382, opened 2021-02-08
Update CMake `Thrust_DIR` documentation.

allisonvacanti:
The cmake packages are not in the CTK, and when we add them they won't
be in the location currently suggested by the docs.


allisonvacanti:
Tangentially related to #1294.

brycelelbach (approved):

PR #1398, opened 2021-03-10
Remove icc from gpuCI.

allisonvacanti:
The prerelease versions of the Intel compilers available freely on ubuntu are too unstable for a CI system.

Disabling the failing CI build until we figure out a long term solution.


allisonvacanti:
run tests

brycelelbach:
Can we instead just mark the failing tests as known failure? We can do the same for NVC++, etc, going forward. It's better to leave the coverage for the tests that are passing instead of loosing the entire config.

PR #1399, opened 2021-03-10
Disable unittest_state_assert on NVC++.

### LINKED ISSUE #1397: Disable thrust.test.unittest_static_assert on NVC++
The machinery for the test framework to hook static assertion failures at runtime uses a `__device__` static global, which NVC++ doesn't support. This causes `thrust.test.unittest_static_assert` to fail. 

We need to either disable this test for NVC++ (preferred because it'll take less time and it is not critical to run this test) or fix it.

@griwes wrote it, so we'll make it his problem.

allisonvacanti:
Fixes #1397.


allisonvacanti:
run tests

brycelelbach (approved):

PR #1424, opened 2021-05-06
fix > intmax num inputs for scan_by_key

### LINKED ISSUE #1410: [BUG] inclusive_scan_by_key OOM on >= INT_MAX elements
Possibly related to the fix at https://github.com/NVIDIA/cub/issues/221, we're seeing an OOM in `inclusive_scan_by_key` when input size is close to (but not exactly) `INT_MAX`.  The limit seems to depend on the accumulator type -- sometimes `(1 << 31) - 1` works, but other times it's `(1 << 31) - 2048` or `(1 << 31) - 4096`.

Tested in `nvidia/cuda:10.2-devel-ubuntu18.04` container with GCC 7.5.0/CUDA 10.2, but present in newer versions too.

```CMake
# CMakeLists.txt
cmake_minimum_required(VERSION 3.20)
project(test VERSION 1.0.0 LANGUAGES C CXX CUDA)
set(CMAKE_EXPORT_COMPILE_COMMANDS ON)

# https://github.com/cpm-cmake/CPM.cmake/blob/310efb9b17d0befe9ccd4f5bf2e39942869777fc/cmake/get_cpm.cmake
set(CPM_DOWNLOAD_VERSION 0.32.0)
if(CPM_SOURCE_CACHE)
  # Expand relative path. This is important if the provided path contains a tilde (~)
  get_filename_component(CPM_SOURCE_CACHE ${CPM_SOURCE_CACHE} ABSOLUTE)
  set(CPM_DOWNLOAD_LOCATION "${CPM_SOURCE_CACHE}/cpm/CPM_${CPM_DOWNLOAD_VERSION}.cmake")
elseif(DEFINED ENV{CPM_SOURCE_CACHE})
  set(CPM_DOWNLOAD_LOCATION "$ENV{CPM_SOURCE_CACHE}/cpm/CPM_${CPM_DOWNLOAD_VERSION}.cmake")
else()
  set(CPM_DOWNLOAD_LOCATION "${CMAKE_BINARY_DIR}/cmake/CPM_${CPM_DOWNLOAD_VERSION}.cmake")
endif()
if(NOT (EXISTS ${CPM_DOWNLOAD_LOCATION}))
  message(STATUS "Downloading CPM.cmake to ${CPM_DOWNLOAD_LOCATION}")
  file(DOWNLOAD
       https://github.com/cpm-cmake/CPM.cmake/releases/download/v${CPM_DOWNLOAD_VERSION}/CPM.cmake
       ${CPM_DOWNLOAD_LOCATION}
  )
endif()
include(${CPM_DOWNLOAD_LOCATION})

# Get Thrust v1.12.0
CPMAddPackage(NAME      Thrust
    VERSION             1.12.0
    GIT_REPOSITORY      https://github.com/NVIDIA/thrust.git
    GIT_TAG             1.12.0
    GIT_SHALLOW         TRUE
    UPDATE_DISCONNECTED TRUE)

thrust_create_target(test::Thrust FROM_OPTIONS)

add_executable(test_inclusive_scan_limit test_inclusive_scan_limit.cu)

target_compile_features(test_inclusive_scan_limit
    PRIVATE cxx_std_14 $<BUILD_INTERFACE:cuda_std_14>)

target_link_libraries(test_inclusive_scan_limit test::Thrust)

execute_process(COMMAND cmake -E create_symlink ${CMAKE_CURRENT_BINARY_DIR}/compile_commands.json  compile_commands.json)
```

```C++
// test_inclusive_scan_limit.cu
#include <thrust/host_vector.h>
#include <thrust/iterator/constant_iterator.h>
#include <thrust/iterator/counting_iterator.h>
#include <thrust/scan.h>

int main(void) {
  //
  // limit is somewhere between these values, otherwise get this error:
  // std::bad_alloc: CUDA error at: rmm/mr/device/cuda_memory_resource.hpp:69:
  // cudaErrorMemoryAllocation out of memory
  //
  auto const scan_by_key_limit = (1uL << 31uL);  // error
  // auto const scan_by_key_limit = (1uL << 31uL) - 1uL;  // no error
  // auto const scan_by_key_limit = (1uL << 31uL) - 2048uL;  // no error
  // auto const scan_by_key_limit = (1uL << 31uL) - 4096uL;  // no error

  auto keys = thrust::make_constant_iterator(0);
  auto vals = thrust::make_counting_iterator(uint64_t{0});

  thrust::host_vector<uint64_t> h_sum(1);
  thrust::device_vector<uint64_t> d_sums(scan_by_key_limit);

  std::cout << "size: " << scan_by_key_limit << std::endl;

  // OOM here
  auto s_end = thrust::inclusive_scan_by_key(
      thrust::device, keys, keys + scan_by_key_limit, vals, d_sums.begin());

  thrust::copy(s_end - 1, s_end, h_sum.begin());

  std::cout << "sums: " << h_sum[0] << std::endl;
}
```

cwharris:
Fixes #1410. With these updates `scan_by_key` supports a higher number of inputs. The number of inputs is now capped by `tile_idx`, which is type `int`. The actual number of supported inputs is `intmax * ITEMS_PER_TILE`, where `ITEMS_PER_TILE` is determined via cub/thrust PtxPolicy.

### DIFF
--- a/thrust/system/cuda/detail/scan_by_key.h
+++ b/thrust/system/cuda/detail/scan_by_key.h
@@ -512,8 +512,8 @@ namespace __scan_by_key {
             inequality_op(equality_op_),
             scan_op(scan_op_)
       {
-        Size tile_idx      = blockIdx.x;
-        Size tile_base     = ITEMS_PER_TILE * tile_idx;
+        int  tile_idx      = blockIdx.x;
+        Size tile_base     = ITEMS_PER_TILE * static_cast<Size>(tile_idx);
         Size num_remaining = num_items - tile_base;
 
         if (num_remaining > ITEMS_PER_TILE)
--- a/thrust/system/cuda/detail/scan_by_key.h
+++ b/thrust/system/cuda/detail/scan_by_key.h
@@ -512,7 +512,7 @@ namespace __scan_by_key {
             inequality_op(equality_op_),
             scan_op(scan_op_)
       {
-        int  tile_idx      = blockIdx.x;
+        Size tile_idx      = blockIdx.x;
         Size tile_base     = ITEMS_PER_TILE * tile_idx;
         Size num_remaining = num_items - tile_base;
 
@@ -734,7 +734,7 @@ namespace __scan_by_key {
                              ScanOp                     scan_op,
                              AddInitToScan              add_init_to_scan)
   {
-    int          num_items    = static_cast<int>(thrust::distance(keys_first, keys_last));
+    Size         num_items    = static_cast<Size>(thrust::distance(keys_first, keys_last));
     size_t       storage_size = 0;
     cudaStream_t stream       = cuda_cub::stream(policy);
     bool         debug_sync   = THRUST_DEBUG_SYNC_FLAG;
--- a/thrust/system/cuda/detail/scan_by_key.h
+++ b/thrust/system/cuda/detail/scan_by_key.h
@@ -734,7 +734,7 @@ namespace __scan_by_key {
                              ScanOp                     scan_op,
                              AddInitToScan              add_init_to_scan)
   {
-    Size         num_items    = static_cast<Size>(thrust::distance(keys_first, keys_last));
+    size_t       num_items    = static_cast<size_t>(thrust::distance(keys_first, keys_last));
     size_t       storage_size = 0;
     cudaStream_t stream       = cuda_cub::stream(policy);
     bool         debug_sync   = THRUST_DEBUG_SYNC_FLAG;


thrust/system/cuda/detail/scan_by_key.h
```diff
@@ -512,7 +512,7 @@ namespace __scan_by_key {
             inequality_op(equality_op_),
             scan_op(scan_op_)
       {
-        int  tile_idx      = blockIdx.x;
+        Size tile_idx      = blockIdx.x;
```
allisonvacanti, line 515: Can you switch this back to using a `static_cast<Size>` in the `tile_base` calculation and leave `tile_idx` as an `int`?

`tile_idx` is passed to `consume_tile`, which expects it to be an `int`, and this change will introduce truncation warnings for 8-byte `Size` types.

allisonvacanti (changes requested):
LGTM aside from one minor change to avoid introducing a build warning.

thrust/system/cuda/detail/scan_by_key.h
cwharris: Ah this is my fault. We should actually we use `Size` for `tile_idx` all the way down, I think. It's conceivable `tile_idx` can rise higher than intmax. In my experiments `ITEMS_PER_TILE was `9 * 256`, meaning `1 >> 43` number of inputs would overflow.

Do you think having `tile_idx` as `Size` would be problematic?

cwharris: I just realized `tile_idx` this goes all the way down to `TilePrefixCallbackOp`, which accepts `int` for `tile_idx` in the constructor, so making a change to `tile_idx`'s type would require changes to all single-pass scan algorithms.

allisonvacanti: A block's x dimension must always fit in an `int`, so it's best to leave `tile_idx` as-is. If we needed more tiles, it'd need to be handled at a higher level of the implementation.

allisonvacanti:
This LGTM, I'll run it through our tests. I should be able to land it before the next release.

Thanks for the patch!

allisonvacanti:
DVS CL: 29947024

run tests

cwharris:
I make the mistake of thinking `Size` was available when determining `num_items`, but that is the point at which `Size` type is determined, so I changed it back to `size_t`.

allisonvacanti (changes requested):
Ah, that's unfortunate. This patch just got a bit more complicated -- see my inline comment.

cwharris:
Closing because this isn't a viable solution without a major overhaul of the single-pass scan utilities, and/or adding conditional dispatched based on the size of input.

allisonvacanti:
We should address this in the long term through NVIDIA/cub#212 and moving Thrust kernels into CUB.

PR #1435, opened 2021-05-13
[gpuCI] Clear stale CMake files and report failures.

allisonvacanti:
I noticed that gcc-5 has been failing for some time, but was reported as passing. This is because the build failures were in CUDA targets, but only non-CUDA tests are run on the `cpu` testers and the builder would only fail if the testing step failed.

Now it checks the configure, build, and test steps explicitly and logs a summary of which steps failed at the end of the output. Now gcc-5 reports as failing:

```
>>>> Summary:
>>>> - Configure Error Code: 0
>>>> - Build Error Code: 1
>>>> - Test Error Code: 0
```

### DIFF
--- a/ci/common/build.bash
+++ b/ci/common/build.bash
@@ -232,7 +232,11 @@ fi
 
 log "Configure Thrust and CUB..."
 
+# Clear out any stale CMake configs:
+rm -rf CMakeCache.txt CMakeFiles/
+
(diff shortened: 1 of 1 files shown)


allisonvacanti:
run tests

allisonvacanti:
@brycelelbach this fixes the build script so that gcc-5 now shows as failing. We were missing these because the tests still passed, even though there were build errors.

Issue #1436, opened 2021-05-17
Too many comparisions during calling to `thrust::sort` and `thrust::stable_sort`


ttnghia:
Please consider the following code snippet:
```
struct counter {
  int* count_ptr;
  __device__ bool operator()(int i, int j) noexcept
  {
    atomicAdd(count_ptr, 1);
    return i < j;
  }
};

rmm::device_uvector<int> indices(2, stream);
thrust::sequence(rmm::exec_policy(stream), indices.begin(), indices.end(), 0);

auto count = cudf::detail::make_zeroed_device_uvector_async<int>(1, stream);
counter _counter{count.begin()};
thrust::sort(rmm::exec_policy(stream), indices.begin(), indices.end(), _counter);

int tmp;
CUDA_TRY(
cudaMemcpyAsync(&tmp, count.begin(), sizeof(int), cudaMemcpyDeviceToHost, stream.value()));
stream.synchronize();

printf("Num. comparisons: %d\n", tmp);
```

In summary, the code above calls `thrust::sort` on a vector of two integer numbers, and counts the number of comparisons. The output of that code is:
```
Num. comparisons: 34817
```

Obviously, such number of comparisons is too excessive when sorting 2 numbers. Since I have only 1 GPU to test (Tesla T4), I don't know if that number is consistent across devices.

I also get the same number of comparisons when using `thrust::stable_sort`.

ttnghia:
Below are some other numbers when sorting arrays with different sizes:
2 --- Num. comparisons: 34817
10 --- Num. comparisons: 34825
100 --- Num. comparisons: 35111
1000 --- Num. comparisons: 40475
10000 --- Num. comparisons: 170980
100000 ---

ttnghia:
I also tried to count the number of comparisons in different ways. For sorting two numbers as above, here is some insight:

- There is exactly 1 comparison of number at index 0 with number at index 1.
- There are 9 comparisons of number at index 1 with number at index 0.
- There are 34680 comparisons of number at index 0 with itself.
- and remaining 127 comparisons of number at index 1 with itself.

senior-zero:
[...]

PR #1442, opened 2021-05-25
Reduce comparisons count in merge sort

senior-zero:
In order to fix [the issue](https://github.com/NVIDIA/thrust/issues/1436) I've reduced the comparisons count in case of partially filled tile. The issue is caused by the fact that each thread has a fixed amount of data to process. If the input data size isn't a multiple of `ITEMS_PER_TILE`, there isn't enough data to fill the local storage of each thread. In this case, extra elements of the storage [are filled with some fixed value](https://github.com/senior-zero/thrust/blob/503806a36518fabf0ace0454deee903aee675614/thrust/system/cuda/detail/sort.h#L491-L494). After that the storage is sorted by `stable_odd_even_sort` by each thread (even ones outside the input range). The idea behind the fix is in skipping `stable_odd_even_sort` for threads of the last tile with no actual input data because in this case their storages are filled with the fixed value. 

The fix doesn't affect performance even for small input data sizes and complex comparators. For input data size of 2 INT32 elements number of comparisons is reduced by 507 times (RTX3090).

### DIFF
--- a/thrust/system/cuda/detail/sort.h
+++ b/thrust/system/cuda/detail/sort.h
@@ -334,6 +334,7 @@ namespace __merge_sort {
       // Parallel thread block merge sort
       //---------------------------------------------------------------------
 
+      template <bool IS_LAST_TILE>
       THRUST_DEVICE_FUNCTION void
       block_mergesort(int tid,
                       int count,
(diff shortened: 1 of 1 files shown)


GPUtester:

allisonvacanti:
add to allowlist

senior-zero:
run tests

thrust/system/cuda/detail/sort.h
```diff
@@ -343,9 +344,10 @@ namespace __merge_sort {
         using core::uninitialized_array;
         using core::sync_threadblock;
 
-        // stable sort items in a single thread
+        // if first element of thread is in input range, stable sort items
         //
-        stable_odd_even_sort(keys_loc,items_loc);
+        if (!IS_LAST_TILE || ITEMS_PER_THREAD * tid < count)
```
allisonvacanti, line 349: Add braces

allisonvacanti (approved):

allisonvacanti:
DVS CL: 30008797

run tests

PR #1492, opened 2021-07-20
Pull the fully qualified cub:: namespace into thrust::

allisonvacanti:
This will allow us to just use `cub::` instead of `CUB_NS_QUALIFIER` as long as we're in `thrust::`.

This effectively reverts a large portion of 363c352 and makes it less obnoxious to use CUB from Thrust.

### DIFF
--- a/thrust/system/cuda/config.h
+++ b/thrust/system/cuda/config.h
@@ -28,6 +28,10 @@
 
 #include <thrust/detail/config.h>
 
+// We don't directly include <cub/version.cuh> since it doesn't exist in
+// older releases. This header will always pull in version info:
+#include <cub/util_namespace.cuh>
+
 #if defined(__CUDACC__) || defined(__NVCOMPILER_CUDA__)
 #  if !defined(__CUDA_ARCH__) || (__CUDA_ARCH__>= 350 && defined(__CUDACC_RDC__))
 #    define __THRUST_HAS_CUDART__ 1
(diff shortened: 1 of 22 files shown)


allisonvacanti:
DVS CL: 30207484

run tests

allisonvacanti:
DVS CL: 30207526

run tests

senior-zero (approved):
Here is the new approach overview:

```cpp
namespace cub {
void g() { }
}

namespace thrust {
namespace cub {
using namespace ::cub;
}
}

namespace cub {
void k() { }
}

namespace thrust {
void f() {
  cub::g();
  cub::k();
}
}

int main() {
  thrust::f();
}
```

I haven't known that it's possible to extend a namespace [after using directive](https://en.cppreference.com/w/cpp/language/namespace#Using-directives). So we don't rely on headers order and have readable code, nice!

allisonvacanti:
It also lets us avoid having to use the `CUB_NS_QUALIFIER` macro from Thrust just to cover the edgecase where `THRUST_WRAPPED_NAMESPACE` and `CUB_WRAPPED_NAMESPACE` are different :)

allisonvacanti:
DVS shows some regressions in cuSPARSE builds with this patch. Investigating...

allisonvacanti:
Added a default definition of the new `CUB_NS_QUALIFIER` macro when `THRUST_IGNORE_CUB_VERSION_CHECK` is defined. Restarting tests:

DVS CL: 30215528

run tests

allisonvacanti:
DVS is still failing, cusparse uses an old copy of CUB that defines `CUB_NS_[PREFIX|POSTFIX]` but not `CUB_NS_QUALIFIER`. I've contacted their maintainers to help resolve this.

allisonvacanti:
Resubmitted DVS under CL 30226678

allisonvacanti:
DVS CL: 30227336

run tests

allisonvacanti:
New DVS CL: 30228004

run tests

allisonvacanti:
The cusparse patch introduced another issue. Working with that team to resolve it before merging.

PR #1499, opened 2021-07-30
Do not commit; testing NVIDIA/cub#163

allisonvacanti:


allisonvacanti:
run tests

allisonvacanti:
gcc 10 failure is due to a known compiler bug in gcc 10.3.

Issue #1214, opened 2020-06-30
`std::allocator` functions and typedefs deprecated in C++17


brycelelbach:
With MSVC 2017:

```
C:/dvs/p4/build/sw/gpgpu/thrust\thrust/detail/allocator/allocator_traits.inl(87): error #1444-D: function "std::allocator<_Ty>::construct(_Objty *, _Types &&...) [with _Ty=vec2, _Objty=vec2, _Types=<>]"
C:/dvs/p4/build/sw/tools/win32/msvc/16.0.0/VC//include\xmemory0(939): here was declared deprecated ("warning STL4010: Various members of std::allocator are deprecated in C++17. Use std::allocator_traits instead of accessing these members directly. You can define _SILENCE_CXX17_OLD_ALLOCATOR_MEMBERS_DEPRECATION_WARNING or _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS to acknowledge that you have received this warning.")
...
```

NVBug 200619442

allisonvacanti:
Reopening -- the typedefs on `std::allocator` were deprecated, too, and are still blocking this for QA.

allisonvacanti:
The root cause of the MSVC/C++17 std::allocator bugs is that CMake will apparently ignore any requested CUDA_STANDARD that it doesn't know how to handle. I've filed a bug about it here: https://gitlab.kitware.com/cmake/cmake/-/issues/20953.

We'll need to bump the CMake version to 3.18 (which is still in RC...) to get support for any C++17 targets.

Fortunately CMake is easy to install on most platforms and the release cycles are short, so we should see a full 3.18 soon. If anyone has issue upgrading in the meantime, let me know.

CMake downloads: https://cmake.org/download/
CMake apt repo instructions: https://apt.kitware.com/

senior-zero:
The problem seems to appear on fresh VS install (16.10.2):

```
C:/Users/evtus/Documents/GitHub/nvidia/thrust\thrust/detail/allocator/allocator_traits.h(143): error C2220: the following warning is treated as an error
[...]

PR #1518, opened 2021-09-07
Fix std::allocator traits

senior-zero:
This PR fixes [the following issue](https://github.com/NVIDIA/thrust/issues/1214). 

SFINAE doesn't allow us to specialize code depending on the deprecation attribute. Therefore existing `allocator_traits` led to deprecated fields accesses. In this PR I explicitly specialized `__THRUST_DEFINE_HAS_NESTED_TYPE` for `std::allocator` in case of C++ >= 17.

### DIFF
--- a/thrust/detail/allocator/allocator_traits.h
+++ b/thrust/detail/allocator/allocator_traits.h
@@ -26,6 +26,8 @@
 #include <thrust/detail/type_traits/has_member_function.h>
 #include <thrust/detail/type_traits.h>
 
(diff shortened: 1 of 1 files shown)


senior-zero:
run tests

senior-zero:
run tests

allisonvacanti:
LGTM, I tested this locally and it seems to work. Just run it through DVS and if it looks good, go ahead and merge it 👍

PR #1519, opened 2021-09-07
Update for `cub::FutureValue` PR (NVIDIA/cub#305)

zasdfgbnm:
Please review https://github.com/NVIDIA/cub/pull/305


allisonvacanti:
@zasdfgbnm Yeah, that's a known compiler bug in gcc 10, it can be ignored.

Internal testing on DVS also looks clean, so this is ready to go in -- thanks for advocating for this and doing the work, I really like how this feature turned out! 😀

I'll get this merged in a moment.

PR #1542, opened 2021-10-21
Avoid include cycles with nvc++ -stdpar

dkolsen-pgi:
Thrust headers cannot include `<memory>` directly.  Doing so may cause compilation errors due to include cycles with `nvc++ -stdpar`.  Headers must include `<thrust/detail/memory_wrapper.h>` instead.

An `#include <memory>` was recently added to thrust/detail/allocator/allocator_traits.h.  Change it to `#include <thrust/detail/memory_wrapper.h>`

PR #1543, opened 2021-10-22
Test that <algorithm> and <memory> aren't included directly.

alliepiper:
These headers cause circular dependencies when Thrust is used as
a backend to nvc++'s stdpar. They've been used accidentally several
times, so it's probably worth automating a check for them.

Expanded and renamed the namespace macro check to test for these headers.


allisonvacanti:
This will need to be rebased once #1542 is merged.

PR #1558, opened 2021-10-30
Optimize merge sort compilation time for C++ < 17

senior-zero:
Merge sort compilation time matched the old version only in C++ 17. It seems that all branches of `doit_step` were compiled in the case of C++ 11 and C++ 14. In this PR, I'm using the cub dispatch layer directly. The compilation times are matching the old version now:

|                  | old | before fix | after fix |
|------------------|-----|------------|-----------|
| c++11 | 7.8 | 12.9       | 7.9       |
| c++14 | 7.9 | 13.1       | 7.8       |
| c++17 | 9.0 | 9.0       | 9.0       |

### DIFF
--- a/thrust/system/cuda/detail/sort.h
+++ b/thrust/system/cuda/detail/sort.h
@@ -55,9 +55,7 @@ namespace cuda_cub {
 
 namespace __merge_sort {
 
-  template <class SORT_ITEMS,
-            class STABLE,
-            class KeysIt,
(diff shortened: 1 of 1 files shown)


senior-zero:
run tests

senior-zero:
DVS: 30598542

allisonvacanti (approved):
Nice! Looks like this was the other missing piece. Let's get this merged today.

PR #1560, opened 2021-11-01
Build bench.cu from CMake.

allisonvacanti:
Builds the older benchmarking executable on non-MSVC platforms when CMake option `THRUST_ENABLE_BENCHMARKS` is enabled. This will be replaced by [thrust_benchmarks](https://github.com/allisonvacanti/thrust_benchmark) before much longer, but we need this to build from CMake in the meantime for Reasons:tm:.

This adds compilation logic only -- It is not executed by `ctest`.

Enabled this for gpuCI/cpu builders, since it compiles fairly quickly.


allisonvacanti:
run tests

senior-zero (approved):

PR #1570, opened 2021-11-19
Add THRUST_PRAGMA_OMP.

allisonvacanti:
Sends all `#pragma omp ...` directives through a new customization point, `THRUST_PRAGMA_OMP(...)`.

Several checks for `THRUST_DEVICE_COMPILER_IS_OMP_CAPABLE` have been removed. These are no longer necessary since the new customization point won't emit any pragmas when OpenMP is disabled.

Fixed a conversion warning in the `sequence.cu` test that started appearing after this change.

cc: @dkolsen-pgi @zkhatami

### DIFF
--- a/testing/sequence.cu
+++ b/testing/sequence.cu
@@ -139,10 +139,21 @@ struct Vector
 };
 
 // Vector-Vector addition
-__host__ __device__ Vector operator+(const Vector a, const Vector b) { return Vector{a.x + b.x, a.y + b.y}; }
+__host__ __device__ Vector operator+(const Vector a, const Vector b)
+{
+  return Vector{a.x + b.x, a.y + b.y};
+}
+
 // Vector-Scalar Multiplication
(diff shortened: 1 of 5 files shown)


allisonvacanti:
run tests

allisonvacanti:
DVS CL: 30678246

dkolsen-pgi:
If a newer version of Thrust which includes this change is used with `nvc++ -stdpar=multicore` version 21.9 or earlier, the `omp_stdpar` pragmas will be ignored and the code will execute sequentially.  I'm fine with that.  I thank that's okay.  NVC++ ships with its own version of Thrust.  The user would have to take extra steps to get the newer version of Thrust on the include path, and that is unlikely to happen when using `-stdpar=multicore`.  I am not proposing that anything be changed because of this; I am just making sure people are aware of the issue.

dkolsen-pgi (approved):

PR #1572, opened 2021-11-30
Reduce header bloat

allisonvacanti:
This PR is part of the if-target split.

The CUDA-specific binary search implementation has been `#ifdef 0`d
for a long time. It didn't perform as well as the sequential
implementation and is dead code that uses old dispatch mechanisms.

This also removes a load of unused headers from
`thrust/system/cuda/execution_policy.h`. The comments around these
headers don't make sense and looks like this was being used for test
bookkeeping.

## Breaking Change

This patch may introduce compile errors in user code when a `thrust::` algorithm is used, but it's header is not included. Such usage of Thrust is not supported and the required headers should be added to resolve any issues.

NVIDIA/cub#407 fixes some related issues in CUB uncovered by this change.


allisonvacanti:
run tests

allisonvacanti:
run tests

senior-zero (approved):

allisonvacanti:
gpuCI failure is unrelated -- I'll fix that timeout in NVIDIA/cub#403.

DVS CL: 30722775

allisonvacanti:
DVS CL: 30723811

run tests

PR #1598, opened 2022-01-21
More gpuCI updates

allisonvacanti:
Requires NVIDIA/cub#425.

This branch makes the following changes to gpuCI:

* Splits `main` integration builds from `prb` pull request builds.
* Split `gpu` builds into `cuda` and `nvhpc` subsets.
* Split `cpu` builds into `cuda` and `nocuda` subsets.

This results in the following 8 new configurations:

* `main-cpu-cuda`
* `main-cpu-nocuda`
* `main-gpu-cuda`
* `main-gpu-nvhpc`
* `prb-cpu-cuda`
* `prb-cpu-nocuda`
* `prb-gpu-cuda`
* `prb-gpu-nvhpc`

Each configuration is controlled by three configuration files:

* an axis definition in `ci/axis/<config>.yml`
* a launch script in `ci/gpuci/<config>.bash`
* build/test configurations in `CMakePresets.json` named `gpuci-<config>`.

The CMake/CTest configuration logic has been moved out of the
`ci/common/build.bash` script and into a `CMakePresets.json` file.
The pre-existing Coverage Plans are still usable from
`ci/local/build.bash`, but each new gpuCI configuration uses its
own unique preset.

Moving all configuration logic to this file will eventually allow us
to manage both gpuCI and DVS configurations from the same place,
using the same building blocks, with greater per-config flexibility.

`ci/common/build.bash` still performs some per-compiler configuration,
but options that describe the amount of work to do has been abstracted
into the presets.


allisonvacanti:
This is blocked waiting for gpuCI to enable the new builds. Once this is done, I'll do some testing and adjust the configs as needed before merging and disabling the old gpuCI builds. I've requested that the old build remain functional during the transition to minimize disruption.

allisonvacanti:
Closing as gpuCI is not the way forward. I'll leave this branch around since the CMakePresets stuff would still be useful at some point, but this version is very out-of-date and will be obsoleted by the monorepo.

PR #1599, opened 2022-01-21
Fix version checks in CMake packages.

allisonvacanti:


allisonvacanti:
run tests

robertmaynard:
This set of changes is insufficient for the following example:

```cmake
list(APPEND CMAKE_PREFIX_PATH <path_to_thrust_1.12> <path_to_thrust_1.15>)
find_package(Thrust REQUIRED)
thrust_create_target(A::Thrust)

find_package(Thrust 1.15 REQUIRED)
thrust_create_target(B::Thrust)
```

The second `find_package(Thrust)` call will occur since we have added a VERSION value, but it will use the cached `find_package(CUB)` call since we don't propagate the requested VERSION argument down.

The naive solution ( as seen below ), resolves this issue:
```
diff --git a/thrust/cmake/thrust-config.cmake b/thrust/cmake/thrust-config.cmake
index 50e84ce7..ae108319 100644
--- a/thrust/cmake/thrust-config.cmake
+++ b/thrust/cmake/thrust-config.cmake
@@ -497,7 +497,8 @@ endfunction()
 macro(_thrust_find_CUDA required)
   if (NOT TARGET Thrust::CUDA)
     thrust_debug("Searching for CUB ${required}" internal)
-    find_package(CUB CONFIG
+    message(STATUS "PACKAGE_FIND_VERSION: ${PACKAGE_FIND_VERSION}")
+    find_package(CUB ${PACKAGE_FIND_VERSION} CONFIG
       ${_THRUST_QUIET_FLAG}
       ${required}
       NO_DEFAULT_PATH # Only check the explicit HINTS below:
```

A full reproducer can be found at: https://gist.github.com/robertmaynard/c6a784e7e78d318cba2ae1e71805c618

allisonvacanti:
Fixed in https://github.com/NVIDIA/thrust/pull/1603

Issue #1478, opened 2021-07-08
Many warnings with -Wconversion


nvibd:
We're seeing a lot of warnings on Linux, host compiler GCC 9.3.0 with -Wconversion.

For a basic example, compile:
```
#include <thrust/device_vector.h>
int main() {
  thrust::device_vector<int> a;
  return 0;
}
```

With: `nvcc main.cu -Xcompiler=-Wconversion`.

This will result in around 1000 lines of output warnings.
Tested with the Thrust versions in CUDA 11.0, 11.3, and the latest commit on main on this repository (`776bbc4d`).

allisonvacanti:
We should add this flag to Thrust and CUB's CI build system for all targets.

PR #1623, opened 2022-02-28
Bug/GitHub/wconversion-1478 Check in

matt-stack:
In reference to https://github.com/NVIDIA/thrust/issues/1478 

Creating PR to get vis on the last remaining tricky type conversions. 

The two pain points are 
1. the conversions required for thrust/system/detail/sequential/scan.h (and similarly for thrust/system/detail/sequential/scan_by_key.h)
2. the 4 spots I detail below 
 
Attaching the list of remaining warnings here:
[conversionWarning47.txt](https://github.com/NVIDIA/thrust/files/8155725/conversionWarning47.txt)

There are four spots where I had a pragma ignore "-Wconversions" because those spots were called so often they would blow up the log, and tricky to solve so I moved on. (Thus they are not present in the log above). I made attempts to create general conversions for them, but was unsuccessfully at the time.

These 4 spots are:
./detail/function.h around line 97
./detail/internal_functional.h around line 332
./detail/tuple.inl (has two spots) around line 479 and 367

I am compiling a log including these errors now, will attach after the fact.

I do not believe I made any changes to the tests themselves, only the source of the type conversions. There are some tests that deliberately test a rotating array of types against each other. 

I believe a trained eye could identify the solutions to the trickier ones that have eluded me.


thrust/system/detail/sequential/scan.h
```diff
@@ -54,6 +54,8 @@ __host__ __device__
 
   // Use the input iterator's value type per https://wg21.link/P0571
   using ValueType = typename thrust::iterator_value<InputIterator>::type;
+  // Use for explicit type conversion
+  using OutputType = typename thrust::iterator_value<OutputIterator>::type;
```
matt-stack, line 58: This was my thinking, that the output type would make sense, but when I do include this I get compile time errors

thrust/iterator/transform_iterator.h
```diff
@@ -312,6 +312,7 @@ template <class AdaptableUnaryFunction, class Iterator, class Reference = use_de
       // convert to their value type before calling m_f. Note that this
       // disallows non-constant operations through m_f.
       typename thrust::iterator_value<Iterator>::type const& x = *this->base();
+      // The issue is that x needs to be the type specified in m_f, but that is unknown 
       return m_f(x);
```
matt-stack, line 316: I tried 
```
return static_cast<typename super_t::reference>(m_f(x));
```
To this one, but I believe it also caused some compile time errors

thrust/system/detail/sequential/scan.h
ericniebler: Output iterators are not required to have a value type.

matt-stack: @ericniebler Thanks, for those that dont have it is it set to null?

thrust/iterator/transform_iterator.h
matt-stack: Looked at this again, and not sure if there is a way (without adding extra args) to inherit the type expected by the unary operator m_f, might have to be ignored from -Wconversion?

thrust/system/detail/sequential/scan.h
matt-stack: Thanks for the tips, I will look into implementing that. There may still be times where when it has to pick the Fallback type (which I am sensing would be ValueType to not break anything), there may still be warning.

PR #1630, opened 2022-03-15
NaN output fix for pow(complex(0,0), 0)

crajnits:
Current implementation outputs NaN for 0^0.
But many widely used formulas involving real-number exponents require 0^0 to be defined as 1.
https://en.wikipedia.org/wiki/Zero_to_the_power_of_zero

### DIFF
--- a/thrust/detail/complex/cpow.h
+++ b/thrust/detail/complex/cpow.h
@@ -39,6 +39,9 @@ complex<typename detail::promoted_numerical_type<T0, T1>::type>
 pow(const complex<T0>& x, const T1& y)
 {
   typedef typename detail::promoted_numerical_type<T0, T1>::type T;
+  if (x.imag() == 0.0) {
+    return pow(x.real(), y);
+  }
   return exp(log(complex<T>(x)) * T(y));
 }
 


GPUtester:

allisonvacanti:
At a quick glance this doesn't seem unreasonable, according to [cppreference](https://en.cppreference.com/w/cpp/numeric/complex/pow):

> The result of `std::pow(0, 0)` is implementation-defined.

This would be a potentially significant change in behavior though, so it might be best for users to implement this on their end (using a custom `pow` function in their library) if they care about getting a specific result from 0^0.

Would implementing this in your code be a reasonable workaround for your usecase @crajnits? Since it's implementation defined, relying on any specific result from this operation will not be portable.

crajnits:
Yes @allisonvacanti , I agree that workaround can be added to the user code itself.
But with current implementation, output of std::pow(0, 0) and thrust::pow(complex(0,0), 0) are not matching.
For example on Windows target, result of std::pow(0, 0) is 1 but thrust::pow(complex(0,0), 0) result is NaN.
In my opinion both result should match.

PR #1631, opened 2022-03-16
Add GDB pretty-printer for thrust vectors and references

### LINKED ISSUE #1318: Provide GDB pretty-printers for Thrust containers and iterators
It would be wonderful if I could inspect the contents of thrust containers: `host_vector` and `device_vector` in GDB (and more importantly, in VSCode).  [GDB allows customizing this](https://www.kurokatta.org/grumble/2018/05/gdb-pretty).

It would save a lot of time if I could inspect device vectors without having to bring them to the host (e.g. the pretty printer script would do that behind the scenes for me). 

Bonus: we also have custom containers with interfaces that look a lot like `device_vector` -- it would be great if Thrust's pretty-printers could be used on those as well.

Bonus 2: Moreover, since fancy iterators are so common in our Thrust code (notably "counting transform" iterators), it could be extremely valuable to be able to introspect the values of iterators. This might require running device code in the general case, but for example, if I want to look at a range of iterators, the pretty-printer might be able to run a `thrust::sequence` on a small range and then copy the results to the host and print them.  Not sure if this is fully supportable, just thought it is a potentially useful idea.

upsj:
I saw #1318 and really liked the idea of providing pretty-printers even for device-side data, so I implemented it in our own library (https://github.com/ginkgo-project/ginkgo/pull/987), and thought it might be useful here as well :) I removed all non-trivial parts of the code that were copied from libstdc++, so there should be no licensing issues. This is ~~should be~~ straightforward to extend to device references ~~, I only had a few issues with getting the device pointer out from a device_reference (gdb returns a heap pointer for `D[0].ptr.m_iterator`, but what looks like a device pointer for `D.m_storage.m_begin.m_iterator.m_iterator` in `examples/basic_vector.cu`, so there must be something weird going wrong with the temporary `D[0]` I guess?)~~

I can't really speak on the potential for hangs that was discussed in the original issue, aside from "it works for me" - that would be something the runtime folks need to answer.

Example run:

```
(gdb) b 33
Breakpoint 1 at 0xc3e8: file /home/tobi/thrust/examples/basic_vector.cu, line 33.
(gdb) run
Starting program: /home/tobi/thrust/build/bin/thrust.example.basic_vector 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
H has size 4
H[0] = 14
H[1] = 20
H[2] = 38
H[3] = 46
H now has size 2
[New Thread 0x7fffefea3000 (LWP 2198)]

Thread 1 "thrust.example." hit Breakpoint 1, main () at /home/tobi/thrust/examples/basic_vector.cu:33
33          D[0] = 99;
(gdb) source ../scripts/gdb-pretty-printers.py 
(gdb) print H
$1 = thrust::host_vector<int, std::allocator<int> > of length 2, capacity 4 = {14, 20}
(gdb) print D
$2 = thrust::device_vector<int, thrust::device_allocator<int> > of length 2, capacity 2 = {14, 20}
(gdb) print D[1]
$3 = (thrust::device_reference<int>) @0xb047f0004: 20
(gdb)
```

Fixes #1318

### DIFF
--- a/scripts/gdb-pretty-printers.py
+++ b/scripts/gdb-pretty-printers.py
@@ -1,0 +1,119 @@
+import gdb
+import sys
+
+if sys.version_info[0] > 2:
+    Iterator = object
+else:
+    # "Polyfill" for Python2 Iterator interface
+    class Iterator:
+        def next(self):
+            return self.__next__()
+
+
+class ThrustVectorPrinter(gdb.printing.PrettyPrinter):
+    "Print a thrust::*_vector"
+
+    class _host_accessible_iterator(Iterator):
+        def __init__(self, start, size):
+            self.item = start
+            self.size = size
+            self.count = 0
+
+        def __iter__(self):
+            return self
+
+        def __next__(self):
+            if self.count >= self.size:
+                raise StopIteration
+            elt = self.item.dereference()
+            count = self.count
+            self.item = self.item + 1
+            self.count = self.count + 1
+            return ('[%d]' % count, elt)
+
+    class _device_iterator(Iterator):
+        def __init__(self, start, size):
+            self.exec = exec
+            self.item = start
+            self.size = size
+            self.count = 0
+            self.buffer = None
+            self.sizeof = self.item.dereference().type.sizeof
+            self.buffer_start = 0
+            # At most 1 MB or size, at least 1
+            self.buffer_size = min(size, max(1, 2 ** 20 // self.sizeof))
+            self.buffer = gdb.parse_and_eval(
+                '(void*)malloc(%s)' % (self.buffer_size * self.sizeof))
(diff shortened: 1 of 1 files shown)


allisonvacanti:
Awesome, thanks! We'll take a look soon, but it may be sometime after GTC.

GPUtester:

upsj:
I figured out the issues with device references, turns out GDB <= 9.2 has some bug I didn't bother to track down that has been fixed in GDB 10.1, so we can actually conditionally add pretty printers for references as well :)

ericniebler (approved):
Looks good to me. Thanks!

PR #1646, opened 2022-03-25
Enable swap(...) for temporaries and between different reference types

### LINKED ISSUE #1645: swap is unnecessarily constrained on reference-like objects
When looking at the implementation of device_reference and tuple_of_iterator_references, I discovered that the `swap` support may be unnecessarily constraining (compared to _BitReference from libstdc++, which I originally took as a reference for how a reference-like type could be implemented):

```cpp
using std::swap;
thrust::device_vector<int> vec(4);
int i{};
auto ref = vec[0];
auto ref2 = vec[1];
swap(vec[0], vec[1]); // 1
swap(ref, ref2);      // 1*
swap(i, vec[0]);      // 2
swap(vec[0], i);      // 3

auto it = thrust::make_zip_iterator(vec.begin(), vec.begin() + 1);
thrust::tuple<int, int> tuple;
swap(it[0], it[2]); // 4
swap(it[0], tuple); // 5
swap(tuple, it[0]); // 6
```

https://godbolt.org/z/4rbK3WhM3

1 does not work (but is intended to work, I guess?) because the swap specialization for `device_reference` unnecessarily (?) requires the input to be a mutable l-value reference instead of just a value, as 1* shows.
2 and 3 do not have associated specializations at all, same as 5 and 6.

This should probably be easy to fix: Add the 4 missing specializations and remove the reference qualifier

upsj:
An attempt at resolving the issues surrounding swap(...) between temporary device references and between different reference(-like) types

Closes #1645

### DIFF
--- a/testing/device_reference.cu
+++ b/testing/device_reference.cu
@@ -208,6 +208,7 @@ DECLARE_UNITTEST(TestDeviceReferenceManipulation);
 
 void TestDeviceReferenceSwap(void)
 {
+  using std::swap;
   typedef int T;
 
   thrust::device_vector<T> v(2);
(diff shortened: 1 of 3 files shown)


GPUtester:

testing/device_reference.cu
```diff
@@ -218,14 +219,37 @@ void TestDeviceReferenceSwap(void)
   ref2 = 13;
 
   // test thrust::swap()
-  thrust::swap(ref1, ref2);
+  swap(ref1, ref2);
```
ericniebler, line 222: I think the point of this code is to test `thrust::swap`, no?

The existence of `thrust::swap` is a bit unfortunate. If you try to `swap` a type that has both `std::` and `thrust::` as associated namespace, the compiler has no way to pick between `std::swap` and `thrust::swap`. We could consider defining `thrust::swap` as a global function _object_ that dispatches to (unqualified) `swap` in a context that has brought `std::swap` in with a `using` declaration. Then a simple call to `thrust::swap` will do the ADL lookup internally, and the potential ambiguity between `std::swap` and `thrust::swap` goes away. I think that's a change that is unlikely to break anybody.

@allisonvacanti thoughts?

ericniebler (commented):
All the thrust reference-like types inherit from `thrust::reference`. We could simplify a lot of this by just defining some hidden `swap` friend functions there.

And that makes me wonder if we want to support arbitrary mixing of reference-like types (e.g., swapping a `device_reference` with kinds of `tagged_reference`). Maybe @jrhemstad could comment on that?

testing/device_reference.cu
allisonvacanti: Good catch, your proposed fix sounds good to me so long as nothing is specializing `thrust::swap`.

The `THRUST_INLINE_CONSTANT` macro may be needed when defining the function object, see https://github.com/NVIDIA/thrust/blob/main/thrust/detail/config/cpp_compatibility.h#L47-L74

PR #1647, opened 2022-03-26
Add __forceinline__ to thrust::detail::wrapped_function::operator()

mkuron:
This pull request adds forced inlining to `thrust::detail::wrapped_function::operator()`. This makes sure that when a function with the `__forceinline__` attribute is wrapped (e.g. when it is passed to `thrust::for_each`), it is actually inlined into the caller (e.g. `thrust::for_each`) and not just inlined into the wrapper, which may or may not have been inlined into the caller automatically.

When a function that does not have the `__forceinline__` attribute is wrapped, this pull request only has a minor effect. Previously the compiler could decide to inline the wrapper into the caller or the function into the wrapper (it would always do one or the other because the wrapper is so simple). Now it can only decide to inline the function into the wrapper or not as the wrapper is always inlined into the caller.

### DIFF
--- a/thrust/detail/function.h
+++ b/thrust/detail/function.h
@@ -42,7 +42,7 @@ struct wrapped_function
 
   __thrust_exec_check_disable__
   template <typename Argument>
-  inline __host__ __device__
+  __thrust_forceinline__ __host__ __device__
   Result operator()(Argument& x) const
   {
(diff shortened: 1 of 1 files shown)


GPUtester:

ericniebler:
run tests

ericniebler (approved):

allisonvacanti:
run tests

PR #1653, opened 2022-03-30
Handle when CMAKE_INSTALL_LIBDIR has nested directories

robertmaynard:


allisonvacanti (approved):

allisonvacanti:
@robertmaynard does libcudacxx need this as well?

Issue #1609, opened 2022-01-27
thrust::reduce_by_key error for 2^31 elements


seunghwak:
Tested environment
GA102 (48 GB GPU memory)
CUDA 11.4
Thrust 1.15.0

The code below failed inside the 'reduce_by_key' call with a memory allocation failure; tried to allocate 18446744073694639872 bytes (16 EB) and failed as my system has only 48 GB. This sounds like a bug related to 32 bit signed integer overflow.

```
cudaDeviceSynchronize();
std::cout << "test thrust reduce_by_key START" << std::endl;
constexpr size_t num_elements = std::size_t{1} << 31;
constexpr int32_t max_key = 8;
thrust::device_vector<int32_t> int_values(num_elements);
thrust::tabulate(thrust::device, int_values.begin(), int_values.end(), [max_key]__device__(auto i) {
  return static_cast<int32_t>(i % max_key);
});
thrust::sort(thrust::device, int_values.begin(), int_values.end());
thrust::device_vector<int32_t> keys(max_key);
thrust::device_vector<size_t> values(max_key);
thrust::reduce_by_key(thrust::device, int_values.begin(), int_values.end(), thrust::make_constant_iterator(size_t{1}), keys.data(), values.data());
cudaDeviceSynchronize();
std::cout << "test thrust reduce_by_key END" << std::endl;
```

senior-zero:
I confirm the issue, here's a reproducer that doesn't require GPU with 48 GB:
```cpp
#include <thrust/device_vector.h>
#include <thrust/tabulate.h>
#include <thrust/reduce.h>
#include <thrust/sort.h>
#include <thrust/iterator/constant_iterator.h>

int main() {
  constexpr size_t num_elements = std::size_t{1} << 31;
  constexpr char max_key = 8;
  thrust::device_vector<char> int_values(num_elements);
  thrust::tabulate(thrust::device, int_values.begin(), int_values.end(),
                   [max_key] __device__(auto i) {
                     return static_cast<char>(i % max_key);
                   });
  thrust::sort(thrust::device, int_values.begin(), int_values.end());
  thrust::device_vector<char> keys(max_key);
  thrust::device_vector<char> values(max_key);
  thrust::reduce_by_key(thrust::device, int_values.begin(), int_values.end(),
                        thrust::make_constant_iterator(char{1}), keys.data(),
                        values.data());
}
```

[...]

PR #1671, opened 2022-04-27
Fix thrust::reduce_by_key for 2^31 elements

senior-zero:
This PR addresses the following [issue](https://github.com/NVIDIA/thrust/issues/1609).

### DIFF
--- a/testing/cuda/reduce_by_key.cu
+++ b/testing/cuda/reduce_by_key.cu
@@ -1,6 +1,11 @@
-#include <unittest/unittest.h>
-#include <thrust/reduce.h>
+#include <thrust/equal.h>
 #include <thrust/execution_policy.h>
+#include <thrust/iterator/counting_iterator.h>
+#include <thrust/iterator/transform_iterator.h>
+#include <thrust/reduce.h>
+#include <unittest/unittest.h>
+
+#include <cstdint>
 
 
 template<typename ExecutionPolicy, typename Iterator1, typename Iterator2, typename Iterator3, typename Iterator4, typename Iterator5>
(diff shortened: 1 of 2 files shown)


senior-zero:
run tests

testing/cuda/reduce_by_key.cu
```diff
@@ -286,3 +289,103 @@ void TestReduceByKeyCudaStreamsNoSync()
 }
 DECLARE_UNITTEST(TestReduceByKeyCudaStreamsNoSync);
 
+
+// Maps indices to key ids
+class div_op : public thrust::unary_function<long long, long long>
```
allisonvacanti, line 294: Consider using `std::int64_t` instead of `long long` to emphasize that this must be a 64-bit type.

testing/cuda/reduce_by_key.cu
```diff
@@ -286,3 +289,103 @@ void TestReduceByKeyCudaStreamsNoSync()
...
+  counting_it count_begin(0ll);
+  counting_it count_end = count_begin + num_items;
+  ASSERT_EQUAL(static_cast<long long>(thrust::distance(count_begin, count_end)),
+               num_items);
+
+  transform_key_it keys_begin(count_begin, div_op{key_size});
+  transform_key_it keys_end(count_end, div_op{key_size});
+
+  transform_val_it values_begin(count_begin, mod_op{key_size});
+  discard_it output_keys;
```
allisonvacanti, line 358: Instead of discarding, we could write an iterator that validates the written keys. For example, see: https://github.com/NVIDIA/thrust/blob/main/testing/async/exclusive_scan/large_indices.cu#L24-L102

allisonvacanti (approved):
Some suggestions inline, but nothing blocking. LGTM.

testing/cuda/reduce_by_key.cu
senior-zero: Thank you for the suggestion! Custom iterator approach leads to quite a lot of code, though. Since there are only 256 unique keys, I'll just store the output keys and compare them with a counting iterator.

senior-zero:
run tests

PR #1683, opened 2022-05-09
Fix bug in final `eval_if` branch in iterator_category_to_traversal.h metafunctions

ericniebler:
Whether this is actually a bug or not depends on intent: if the intent is to force a hard error if we reach the final `eval_if` branch, then the code is correct as-is. If, however, the intent is for the metafunctions to evaluate to `void` as I suspect, then this is a bug.


ericniebler:
run tests

allisonvacanti (approved):
LGTM, agreed that the intent was likely to alias void rather than error out.

Issue #1205, opened 2020-06-25
testing/regression/* not used by CMake/CTest


allisonvacanti:
Naively trying to add them cause issues on windows due to CMAKE_OBJECT_PATH_MAX -- the paths to the object files exceed a system-dependent maximum number of characters. We should find a way to add these. 

We could shorten the names, but it might be best to just integrate them with the other unit tests in `testing/` or `testing/<backend>` and put bug metadata in comments.

allisonvacanti:
Just noticed that some of these aren't meant to be executed and just check that compilation succeeds, e.g:

https://github.com/NVIDIA/thrust/blob/main/testing/regression/gh_911__merge_by_key_wrong_element_type_default_comparator.cu

I can't find an existing build system for these, so I'm not sure which are meant to be run (if any). @brycelelbach do you recall if these are all intended to be compile only tests?

allisonvacanti:
Just noticed that some of these aren't meant to be executed and just check that compilation succeeds, e.g:

https://github.com/NVIDIA/thrust/blob/main/testing/regression/gh_911__merge_by_key_wrong_element_type_default_comparator.cu

I can't find an existing build system for these, so I'm not sure which are meant to be run (if any). @brycelelbach do you recall if these are all intended to be compile only tests?

brycelelbach:
Some of them are, some are meant to be run. It was an idea that never really came to fruition.

allisonvacanti:
We should incorporate the ones that run into the existing unit tests and remove this directory, since it's experimental / unused.

allisonvacanti:
There's some overlap here with #1670 -- once we have proper compile test support we could use that to check the "expected build failure" tests.

PR #1685, opened 2022-05-10
partial fix for iterator traits and tags

ericniebler:
The symptoms of this problem are described in #902. Really, there are two problems with the way Thrust's iterator tags are currently defined:

1. The `*_device_*` iterator tags are erroneously convertible the `*_host_*` tags. That's because the `*_host_*` tags are simply type aliases for the standard iterator tags, which the `*_device_*` tags inherit from.

2. Because of the way the device iterator tags are defined (with `iterator_category_with_system_and_traversal`) there is no relationship between the different device tags. We would expect (and much of the code does) that `thrust::forward_device_iterator_tag` is convertible to `thrust::input_device_iterator_tag`, but that is sadly not true.

The two traits, `iterator_category_to_system` and `iterator_category_to_traversal`, are first checking whether the category is convertible to a host iterator tag. That is trivially true even for the device tags because of (1); as a result, device tags appear to have the "host" system, which is incorrect.

This PR offers the simplest, least intrusive, and sadly partial fix to the problem reported in #902. It changes the two traits so that they first check for convertibility to the _device_ tag. This gets the answer correctly in more cases than currently.

I _think_ I know how to fix this properly, but not without potentially breaking some code, so I'm submitting this PR now as a stop-gap.

### DIFF
--- a/testing/regression/CMakeLists.txt
+++ b/testing/regression/CMakeLists.txt
@@ -1,3 +1,7 @@
+if ("thrust" IN_LIST THRUST_TARGETS)
+  thrust_add_test(test_target "regression.gh_902__iterator_category_to_system_returns_wrong_tag" "gh_902__iterator_category_to_system_returns_wrong_tag.cu" "thrust")
+endif()
+
 #
 # Disabled as these test names are too long for CMAKE_OBJECT_PATH_MAX.
 # We should integrate these with the other unit tests.
--- a/testing/regression/gh_902__iterator_category_to_system_returns_wrong_tag.cu
+++ b/testing/regression/gh_902__iterator_category_to_system_returns_wrong_tag.cu
@@ -1,0 +1,22 @@
+#include <thrust/iterator/detail/iterator_category_with_system_and_traversal.h>
+#include <thrust/iterator/iterator_categories.h>
+#include <thrust/iterator/iterator_traits.h>
+
+template <class IteratorCategory>
+using _category_to_system_t =
+    typename thrust::detail::iterator_category_to_system<IteratorCategory>::type;
+
+static_assert(std::is_same<thrust::device_system_tag, _category_to_system_t<thrust::output_device_iterator_tag>>::value);
+static_assert(std::is_same<thrust::device_system_tag, _category_to_system_t<thrust::input_device_iterator_tag>>::value);
+// static_assert(std::is_same<thrust::device_system_tag, _category_to_system_t<thrust::forward_device_iterator_tag>>::value);
+// static_assert(std::is_same<thrust::device_system_tag, _category_to_system_t<thrust::bidirectional_device_iterator_tag>>::value);
+// static_assert(std::is_same<thrust::device_system_tag, _category_to_system_t<thrust::random_access_device_iterator_tag>>::value);
+
+static_assert(std::is_same<thrust::host_system_tag, _category_to_system_t<thrust::output_host_iterator_tag>>::value);
+static_assert(std::is_same<thrust::host_system_tag, _category_to_system_t<thrust::input_host_iterator_tag>>::value);
+static_assert(std::is_same<thrust::host_system_tag, _category_to_system_t<thrust::forward_host_iterator_tag>>::value);
+static_assert(std::is_same<thrust::host_system_tag, _category_to_system_t<thrust::bidirectional_host_iterator_tag>>::value);
+static_assert(std::is_same<thrust::host_system_tag, _category_to_system_t<thrust::random_access_host_iterator_tag>>::value);
+
+// static_assert(!std::is_convertible<thrust::input_device_iterator_tag, thrust::input_host_iterator_tag>::value);
+static_assert(!std::is_convertible<thrust::input_host_iterator_tag, thrust::input_device_iterator_tag>::value);
--- a/thrust/iterator/detail/iterator_category_to_system.h
+++ b/thrust/iterator/detail/iterator_category_to_system.h
@@ -38,24 +38,24 @@ template <typename> struct device_iterator_category_to_backend_system;
 // we should just specialize this metafunction for iterator_category_with_system_and_traversal
 template<typename Category>
   struct iterator_category_to_system
-    // convertible to host iterator?
+    // convertible to device iterator?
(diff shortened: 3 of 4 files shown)


ericniebler:
run tests

testing/regression/CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
+if ("thrust" IN_LIST THRUST_TARGETS)
```
allisonvacanti, line 1: We'll need to move this test into `testing/` and shorten its name (probably just `iterator_categories.cu`). The tests in the `testing/regression` directory are problematic and need to be cleaned up. See #1205 for details.

testing/regression/gh_902__iterator_category_to_system_returns_wrong_tag.cu
```diff
@@ -0,0 +1,22 @@
...
+#include <thrust/iterator/iterator_categories.h>
+#include <thrust/iterator/iterator_traits.h>
+
+template <class IteratorCategory>
+using _category_to_system_t =
+    typename thrust::detail::iterator_category_to_system<IteratorCategory>::type;
+
+static_assert(std::is_same<thrust::device_system_tag, _category_to_system_t<thrust::output_device_iterator_tag>>::value);
+static_assert(std::is_same<thrust::device_system_tag, _category_to_system_t<thrust::input_device_iterator_tag>>::value);
+// static_assert(std::is_same<thrust::device_system_tag, _category_to_system_t<thrust::forward_device_iterator_tag>>::value);
```
allisonvacanti, line 11: Are the commented out examples still broken? Let's either enable them or add a comment explaining why they're disabled.

testing/regression/gh_902__iterator_category_to_system_returns_wrong_tag.cu
```diff
@@ -0,0 +1,22 @@
+#include <thrust/iterator/detail/iterator_category_with_system_and_traversal.h>
+#include <thrust/iterator/iterator_categories.h>
+#include <thrust/iterator/iterator_traits.h>
+
+template <class IteratorCategory>
+using _category_to_system_t =
+    typename thrust::detail::iterator_category_to_system<IteratorCategory>::type;
+
+static_assert(std::is_same<thrust::device_system_tag, _category_to_system_t<thrust::output_device_iterator_tag>>::value);
```
allisonvacanti, line 9: We need to build with C++11 + C++14, so these will need to use the `static_assert(cond, msg)` syntax.

allisonvacanti: Looks like CI didn't catch this because the tests is only added when the `thrust` target is defined in `THRUST_TARGETS`. For multiconfig builds (used by CI), this target won't exist and we'll instead have `thrust.[host].[device].[dialect]` targets.

ericniebler:
@allisonvacanti I see you tagged this with "2.0" milestone, but this is a change that is safe to make in the 1.X series, IMO.

allisonvacanti:
> this is a change that is safe to make in the 1.X series, IMO.

I would be, but the RC is already out. We only restart the RC process to add fixes for regressions.

ericniebler:
run tests

ericniebler:
I suppose if an iterator had `any_system_tag` as its system tag, then this diff would change its reported system from host to device. :thinking:

allisonvacanti:
> do any iterators use any_system_tag?

Yes -- this tag is used by iterators that aren't backed by actual memory (e.g `thrust::counting_iterator`, `thrust::constant_iterator`, etc, as well as several in CUB).

> I suppose if an iterator had any_system_tag as its system tag, then this diff would change its reported system from host to device. 🤔

This should be fine. The device system is typically faster than the host system, so this is the preferred behavior. If an algorithm takes an `any_system` iterator and also a `host_system` iterator (e.g. `thrust::copy(any_system_it, any_system_it_end, host_system_it_begin)`) and requires the algorithm to run on host, it should still work since we check the systems of all iterators while dispatching.

allisonvacanti (approved):

miscco:
@ericniebler @allisonvacanti 

Is this ready to merge?

ericniebler:
> Is this ready to merge?

I ... think so? I can rebase, retest and commit if all looks good.

ericniebler:
run tests

miscco:
Hm that one is strange:
```
/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/ext/numeric_traits.h(70): error: qualified name is not allowed

/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/ext/numeric_traits.h(72): error: expected a "("

/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/ext/numeric_traits.h(72): error: expected a type specifier

/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/ext/numeric_traits.h(72): error: expected a ")"

/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/ext/numeric_traits.h(76): error: expected a "("

/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/ext/numeric_traits.h(76): error: expected a type specifier

/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/ext/numeric_traits.h(77): error: expected a ";"

/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/ext/numeric_traits.h(78): error: expected a "("

/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/ext/numeric_traits.h(78): error: expected a ";"

/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/ext/numeric_traits.h(88): error: class template "__gnu_cxx::__numeric_traits_integer<_Value>" has no member "__is_signed"
```

There is nothing in this PR that could have triggered this

senior-zero:
@miscco we have wait till https://github.com/NVIDIA/thrust/pull/1866 is merged.

PR #1687, opened 2022-05-12
Make any_system_tag only convertible to other system tags

ericniebler:
`any_system_tag` is currently defined as:

```c++
struct any_system_tag
  : thrust::execution_policy<any_system_tag>
{
  // allow any_system_tag to convert to any type at all
  // XXX make this safer using enable_if<is_tag<T>> upon c++11
  template<typename T> operator T () const {return T();}
};
```

We're C++11 now, so we can and should constrain this conversion operator.

This change is relevant to the future work on iterator categories and tag dispatching.

### DIFF
--- a/thrust/iterator/detail/any_system_tag.h
+++ b/thrust/iterator/detail/any_system_tag.h
@@ -18,6 +18,7 @@
 
 #include <thrust/detail/config.h>
 #include <thrust/detail/execution_policy.h>
+#include <thrust/detail/type_traits.h>
 
 THRUST_NAMESPACE_BEGIN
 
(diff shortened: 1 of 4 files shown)


ericniebler:
run tests

thrust/iterator/detail/any_system_tag.h
```diff
@@ -18,16 +18,28 @@
...
 #include <thrust/detail/config.h>
 #include <thrust/detail/execution_policy.h>
+#include <thrust/detail/type_traits.h>
 
 THRUST_NAMESPACE_BEGIN
 
 struct any_system_tag
   : thrust::execution_policy<any_system_tag>
 {
   // allow any_system_tag to convert to any type at all
```
allisonvacanti, line 28: nit: 
```suggestion
  // allow any_system_tag to convert to any system tag type
```

allisonvacanti (approved):

PR #1691, opened 2022-05-16
Don't specialize is_system_tag for host_system_tag when compiling for device

ericniebler:


ericniebler:
run tests

allisonvacanti (approved):
We should add a comment explaining why this specialization is disabled, but LGTM.

allisonvacanti:
Looks like there are some new failures that will need to be addressed before merging.

PR #1696, opened 2022-05-19
add tests for scan_by_key in-place execution

senior-zero:
Our scan by key implementation doesn't allow keys/result aliasing. This is caused by non-synchronized reading of `d_keys_in[tile_base - 1]` in the [following code](https://github.com/NVIDIA/cub/blob/f80aa78d9d1fbce45e1ec7293952131ef2e31286/cub/agent/agent_scan_by_key.cuh#L367). Therefore, aliasing of keys and results introduces a data race. The following code can be used as a reproducer:
```cpp
#include <thrust/scan.h>

int main() {
  const std::size_t n = 64 * 1024 * 1024;
  thrust::device_vector<int> keys(n, 1);
  thrust::device_vector<int> values(n, 1);
  thrust::device_vector<int> output_storage(n);

  // thrust::device_vector<int> &output = values; // alias is OK
  thrust::device_vector<int> &output = keys; // alias is NOT OK

  thrust::exclusive_scan_by_key(keys.begin(), keys.end(), values.begin(), output.begin());

  if (output[n - 1] != n - 1) {
    std::cerr << "wrong result: " << output[n - 1] << " != " << n << std::endl;
  }
}
```

We also don't have tests for this use case. I can fix the code to guarantee this behaviour (just like in adjacent difference), but it'll slow down non-in-place execution in some cases. My suggestion is to have a quick fix of documentation.

### DIFF
--- a/testing/cuda/scan_by_key.cu
+++ b/testing/cuda/scan_by_key.cu
@@ -78,7 +78,7 @@ void TestScanByKeyDevice(ExecutionPolicy exec)
   }
   ASSERT_EQUAL(d_output, h_output);
   
-  // in-place scans
+  // in-place scans: in/out values aliasing
   h_output = h_vals;
   d_output = d_vals;
   thrust::inclusive_scan_by_key(h_keys.begin(), h_keys.end(), h_output.begin(), h_output.begin());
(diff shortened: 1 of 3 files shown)


allisonvacanti:
Summarizing some offline discussions and testing:

- This has been broken for years, and is not a regression introduced by the recent changes to the `scan_by_key` implementation.
- The common case of overlapping `values_in`/`values_out` is fine, the only issue is overlapping `keys_in`/`values_out`, which is a somewhat odd usecase.
- Based on similar situations, fixing this is expected to introduce ~15% slowdown.

@senior-zero will be investigating this a bit more before we make a decision here.

senior-zero:
I've managed to fuse some kernels to reduce slowdown up to 6% on small to moderate problem sizes. The performance regression disappears at about `2^20` items for `std::uint{16,32,64}_t` and at about `2^22` for `std::uint8_t`.

senior-zero:
@allisonvacanti I've bumped CUB version to one that incorporates updated scan by key. I'd like to re purpose this PR to update tests in thrust. Please, review the changes.

senior-zero:
run tests

allisonvacanti (changes requested):
Let's also add tests to the `thrust/testing/scan_by_key.*.cu` files to make sure the non-CUDA backends test this, too.

senior-zero:
run tests

allisonvacanti (approved):

testing/cuda/scan_by_key.cu
```diff
@@ -98,6 +98,24 @@ void TestScanByKeyDevice(ExecutionPolicy exec)
...
   ASSERT_EQUAL(d_output, h_output);
+
+  // in-place scans: keys/values aliasing
+  thrust::inclusive_scan_by_key(h_keys.begin(), h_keys.end(), h_vals.begin(), h_output.begin());
+  inclusive_scan_by_key_kernel<<<1,1>>>(exec, d_keys.begin(), d_keys.end(), d_vals.begin(), d_keys.begin());
+  {
+    cudaError_t const err = cudaDeviceSynchronize();
+    ASSERT_EQUAL(cudaSuccess, err);
+  }
+  ASSERT_EQUAL(d_keys, h_output);
```
miscco, line 109: Nitpick, no change requested: I would prefer if we put the whole test in its own block rather than only the check, as that helps readability a lot
```suggestion
  { // in-place scans: keys/values aliasing
    thrust::inclusive_scan_by_key(h_keys.begin(), h_keys.end(), h_vals.begin(), h_output.begin());
    inclusive_scan_by_key_kernel<<<1,1>>>(exec, d_keys.begin(), d_keys.end(), d_vals.begin(), d_keys.begin());
    cudaError_t const err = cudaDeviceSynchronize();
    ASSERT_EQUAL(cudaSuccess, err);
    ASSERT_EQUAL(d_keys, h_output);
  }
```

PR #1744, opened 2022-07-17
Squelch a bunch of warnings coming from thrust headers.

tomilov:
If something is compiled having `THRUST_DEVICE_SYSTEM=CUDA` set, then user get a tons of warnings coming from `.inl` files. All of them looks like `unused function [-Wunused-function]`.

### DIFF
--- a/thrust/cmake/thrust-config.cmake
+++ b/thrust/cmake/thrust-config.cmake
@@ -684,7 +684,7 @@ if (NOT TARGET Thrust::Thrust)
     FORCE
   )
   unset(_THRUST_VERSION_INCLUDE_DIR CACHE) # Clear tmp variable from cache
-  target_include_directories(_Thrust_Thrust INTERFACE "${_THRUST_INCLUDE_DIR}")
+  target_include_directories(_Thrust_Thrust SYSTEM INTERFACE "${_THRUST_INCLUDE_DIR}")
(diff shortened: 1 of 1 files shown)


GPUtester:

miscco:
I am really not a fan of this nuclear option. I would rather add specific suppression directives for each warning that is not fixable / not worth it and fix the others

allisonvacanti:
You're absolutely correct that these **should** be system includes when using that target. Ideally, the targets generated by our packages would be proper `IMPORTED` targets, which generate system include paths by default.

Unfortunately, we can't do this because of how nvcc injects the CUDA Toolkit include path during compilation. See [this bit](https://github.com/NVIDIA/thrust/pull/1744/files#diff-c7efc25e141c19e87e995d58d4c8e3c51d6d9a291146d4e3a409b5c2a1ae9ab8L378-L390) for details, but the short version is that this change would cause nvcc to completely ignore the Thrust you're trying to use and instead use the Thrust headers from the CUDA toolkit.

This issue with user warnings is something we're painfully aware of. If you'd like to file an issue with the specific warning flags you care about, we can prioritize fixing/suppressing them in a future release.

robertmaynard:
You can track my work on getting CMake to always mark the CUDA Toolkit include as system here: https://gitlab.kitware.com/cmake/cmake/-/issues/23731

That would allow us to re-open this PR in the future.

PR #1751, opened 2022-07-21
Fix RDC flags on nvc++ builds.

allisonvacanti:
nvcc defaults to rdc-off, nvc++ defaults to rdc-on. We need to explicitly
enable or disable these flags for each CUDA target, rather than just
enabling them when needed.

### DIFF
--- a/cmake/ThrustUtilities.cmake
+++ b/cmake/ThrustUtilities.cmake
@@ -11,15 +11,22 @@ function(thrust_wrap_cu_in_cpp cpp_file_var cu_file thrust_target)
   set(${cpp_file_var} "${cpp_file}" PARENT_SCOPE)
 endfunction()
 
-# Enable RDC for a CUDA target. Encapsulates compiler hacks:
-function(thrust_enable_rdc_for_cuda_target target_name)
+# Enable or disable RDC for a CUDA target.
+# Just using the CMake property won't work for our nvcxx builds, we need
+# to manually specify flags.
+# nvcc disables RDC by default, while nvc++ enables it. Thus this function
+# must be called on all CUDA targets to get consistent RDC state across all
+# platforms.
+function(thrust_set_rdc_state target_name enable)
   if ("NVCXX" STREQUAL "${CMAKE_CUDA_COMPILER_ID}")
-    set_target_properties(${target_name} PROPERTIES
-      COMPILE_FLAGS "-gpu=rdc"
-    )
+    if (enable)
+      target_compile_options(${target_name} PRIVATE "-gpu=rdc")
+    else()
+      target_compile_options(${target_name} PRIVATE "-gpu=nordc")
+    endif()
   else()
     set_target_properties(${target_name} PROPERTIES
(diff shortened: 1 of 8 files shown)


allisonvacanti:
run tests

internal/benchmark/CMakeLists.txt
```diff
@@ -24,6 +24,10 @@ foreach(thrust_target IN LISTS THRUST_TARGETS)
   target_include_directories(${bench_target} PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}")
   thrust_clone_target_properties(${bench_target} ${thrust_target})
 
+  if ("CUDA" STREQUAL "${config_device}")
+    thrust_set_rdc_state(${example_target} OFF)
```
robertmaynard, line 28: no benchmark version of `THRUST_ENABLE_TESTS_WITH_RDC`?

cmake/ThrustUtilities.cmake
```diff
@@ -11,15 +11,22 @@ function(thrust_wrap_cu_in_cpp cpp_file_var cu_file thrust_target)
...
 
-# Enable RDC for a CUDA target. Encapsulates compiler hacks:
-function(thrust_enable_rdc_for_cuda_target target_name)
+# Enable or disable RDC for a CUDA target.
+# Just using the CMake property won't work for our nvcxx builds, we need
+# to manually specify flags.
+# nvcc disables RDC by default, while nvc++ enables it. Thus this function
+# must be called on all CUDA targets to get consistent RDC state across all
+# platforms.
+function(thrust_set_rdc_state target_name enable)
```
robertmaynard, line 20: Any reason we don't roll the `if("CUDA" STREQUAL "${config_device})" in to this function instead of having to guard each call?

internal/benchmark/CMakeLists.txt
allisonvacanti: Good question -- nope, this is a legacy benchmark that is planned to be replaced by the NVBench suite. The build system for it is pretty barebones and just emulates old behaviors.

cmake/ThrustUtilities.cmake
allisonvacanti: This would simplify things a bit, but setting RDC state on non-CUDA targets wouldn't make sense so I wanted to avoid using this on non-CUDA targets.

senior-zero (approved):

dkolsen-pgi:
I understand the problem and agree with the goal of this.  But I don't know CMake well enough to review the actual changes in any useful way.

miscco (approved):

allisonvacanti:
run tests

allisonvacanti:
Looks like this change exposed some bugs in nvc++ now that some translation units turn RDC off. I'll aim to reduce these and file some bug reports next week.

allisonvacanti:
@miscco is looking into the nvc++ failures.

allisonvacanti:
This will no longer be needed after an upcoming CMake refactoring.

PR #1756, opened 2022-07-27
copy_if doesn't require copy assignment operator for OutputIterator

mfbalin:
For OutputIterators who don't have copy assignment operator, the assignment into result was giving a compiler error. This change avoids using the copy assignment operator.

### DIFF
--- a/thrust/system/cuda/detail/cdp_dispatch.h
+++ b/thrust/system/cuda/detail/cdp_dispatch.h
@@ -46,7 +46,7 @@
 
 // seq_impl unused.
 #define THRUST_CDP_DISPATCH(par_impl, seq_impl)                                \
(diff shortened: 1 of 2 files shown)


GPUtester:

allisonvacanti (changes requested):
The change looks reasonable, but go ahead and remove the `return` statement following the `THRUST_CDP_DISPATCH` macro -- compilers will flag that as unreachable code and emit warnings.

allisonvacanti (approved):

allisonvacanti:
run tests

miscco (approved):
Thanks a lot for that contribution! 🎉

allisonvacanti:
run tests

PR #1765, opened 2022-08-12
Build docs from new commits to main.

bdice:
The thrust docs have fallen out of date because the GitHub Action that builds them is targeting the wrong branch. This PR fixes that so that the docs are built from `main` on every commit. I also added a manual trigger that will appear as a button on the following page: https://github.com/NVIDIA/thrust/actions/workflows/deploy-documentation-github-pages.yml

### DIFF
--- a/.github/workflows/deploy-documentation-github-pages.yml
+++ b/.github/workflows/deploy-documentation-github-pages.yml
@@ -3,7 +3,10 @@ name: Deploy Documentation GitHub Pages
 on:
   push:
     branches:
-      - feature/new-docs
+      - "main"
+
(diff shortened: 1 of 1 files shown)


allisonvacanti:
LGTM, let's try it out. @bdice can you check that it works as expected and let us know if there's an issue?

allisonvacanti (approved):

bdice:
@allisonvacanti Yes! This action worked as expected. https://nvidia.github.io/thrust/releases.html now shows the latest changelog (and my previous fix to the search bar in #1644). Thanks for your help. 🚀

PR #1777, opened 2022-08-26
Fix building against libc++ on Windows

georgthegreat:
In microsoft/stl `std::_Unwrapped_t` [is just](https://github.com/microsoft/STL/blob/2a1b881e2fae9c42026040a7fdfc5a93e59a91c8/stl/inc/xutility#L935) `std::remove_cvref_t`.


GPUtester:

georgthegreat:
So, how do I make thrust compile againt different STL on Windows?
I can check microsoft/stl specific define if it has one.

miscco:
That depends a lot on what you want to do. What VS version do you have, what CUDA version etc. (AFAIK  11.6 is needed for current MSVC STL)

PR #1779, opened 2022-08-29
Extend clang-format file to sort the includes by category

miscco:


senior-zero (approved):
You are my hero :smile: Could you port this to CUB and [nvbench](https://github.com/NVIDIA/nvbench) as well?

PR #1797, opened 2022-09-19
Add a `THRUST_HOST_DEVICE` macro

miscco:
clang-format is generally confused by our `__host__` and `__device__` macros, severely messing up formatting.

We can replace that by a single macro and tell clang-format to not mess with its formatting.

While we are at it add some other common macros to the list of `StatementMacros`


.clang-format
```diff
@@ -86,5 +86,14 @@ SpacesInCStyleCastParentheses: false
 SpacesInParentheses: false
 SpacesInSquareBrackets: false
 Standard: c++11
+StatementMacros:
```
miscco, line 89: Those are the relevant changes.

It tells clang-format to treat the macro like a statement, so that it adds a linebreak without indentation afterwards

.clang-format
```diff
@@ -86,5 +86,14 @@ SpacesInCStyleCastParentheses: false
...
+  - __thrust_exec_check_disable__
+  - __host__
+  - __device__
```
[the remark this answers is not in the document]
miscco: From my point of view we should add all those macros

thrust/detail/config/host_device.h
```diff
@@ -42,3 +42,9 @@
 
 #endif
 
+/**
+ * \def THRUST_HOST_DEVICE
+ * This macro is used replace __host__ __device__, with a clang-format friendly
+ * replacement
+ */
+#define THRUST_HOST_DEVICE __host__ __device__
```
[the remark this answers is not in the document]
miscco: From my point of view it matters a lot, as it greatly enlarges function definitions. 

If we have the convention that all combinations of {`__host__`, `__device__`} are on the same line before the function declaration, it greatly reduces mental overhead

Actually I am really in favor of pushing all those definitions into libcu++ and have a single source of truth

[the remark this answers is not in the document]
allisonvacanti: My 2c, I prefer having the attributes / annotations on a separate line from the function definition. It makes it much easier to parse the function def at a glance, since the return type is always as the beginning of the line. Having to parse a variable number of tokens from the start of a function def adds extra steps when skimming code.

allisonvacanti:
Interesting -- this has been an annoyance for so long!

I'm curious why we'd need the macro. Why doesn't adding `__host__ __device__` do the same thing as just `__host__` or `__device__`?

.clang-format
```diff
@@ -67,7 +67,7 @@ PenaltyBreakString: 70
 PenaltyBreakTemplateDeclaration: 0
 PenaltyExcessCharacter: 100
 PenaltyReturnTypeOnItsOwnLine: 90
-PointerAlignment: Right
+PointerAlignment: Left
```
allisonvacanti, line 70: This should be a different discussion. One bikeshed at a time please :)

```suggestion
PointerAlignment: Right
```

.clang-format
```diff
@@ -86,5 +86,18 @@ SpacesInCStyleCastParentheses: false
...
+StatementMacros:
+  - __thrust_exec_check_disable__
+  - __host__
+  - __device__
+  - __global__
+  - THRUST_DEVICE_FUNCTION
+  - THRUST_HOST_FUNCTION
+  - THRUST_FUNCTION
+  - THRUST_HOST_DEVICE
+  - THRUST_NAMESPACE_BEGIN
```
allisonvacanti, line 98: `NamespaceMacros` or `MacroBlock[Begin|End]` seem more relevant for these ([link](https://clang.llvm.org/docs/ClangFormatStyleOptions.html)).

.clang-format
```diff
@@ -86,5 +86,18 @@ SpacesInCStyleCastParentheses: false
 SpacesInParentheses: false
 SpacesInSquareBrackets: false
 Standard: c++11
+StatementMacros:
```
allisonvacanti, line 89: Reviewing the different `Macro` options in the clang-format docs, it looks like they did add an `AttributeMacros` block. This seems more appropriate for these ([link](https://clang.llvm.org/docs/ClangFormatStyleOptions.html)).

.clang-format
miscco: > Interesting -- this has been an annoyance for so long!
> 
> I'm curious why we'd need the macro. If `__host__` and `__device__` are both marked as statement macros, `__host__ __device__` should work too, right?

No, because we tell `clang-format` that `__host__` is a statement so it will change

allisonvacanti: Ouch, yeah, that's rough. Maybe `AttributeMacros` would behave better? I'd like to avoid adding a special macro here if possible.

miscco:
I am closing this for now. It would make more sense to find a common style for all three repositories

PR #1799, opened 2022-09-21
Remove remnants of `throw()`

miscco:
the `throw()` specification has been removed with C++20 and will error out on us.

So rather than that, simply use noexcept, as C++03 is thankfully a thing of the past

Fixes nvbug3799847


miscco:
run tests

senior-zero (approved):
Thanks for the cleanup!

dkolsen-pgi (approved):

PR #1825, opened 2022-11-08
Correct install(PATTERN) to match end of filenames

robertmaynard:
The `PATTERN` matches at the end of filenames not the front, so we need to specify `*.cmake.in` to exclude the template files.


senior-zero (changes requested):
The same comment as on the CUB side of this PR. `*.cmake.in` files are used below to generate `*.cmake` files on the installation side. I don't see why we need `*.cmake.in` files in the installation. Unless I'm missing something, let's remove this.

senior-zero (approved):

PR #1832, opened 2022-11-25
Let CUB select reduce offsets

senior-zero:
Found a bug in device-side thrust reduce, which is described [here](https://github.com/NVIDIA/cub/pull/597). This PR adds a test for device-side reduction of large number of items and also defers offset selection to CUB.


senior-zero:
run tests

miscco (approved):

PR #1848, opened 2023-01-13
Thrust: providing the error messages about the lack of GPU or a GPU w…

zkhatami:
To give a user some clue what's happening if the program gets compiled on a node with no GPU or if it gets compiled with different compute capability than the one it's running on. In both scenarios no good error message was produced before. The proposed changes will improve the user experience and make it easier for users to troubleshoot problems.

This fix is for addressing the issue#1785 reported on Thrust https://github.com/NVIDIA/thrust/issues/1785

From issue#1785 on thrust (https://github.com/NVIDIA/thrust/issues/1785), for this small test case:

`
#include <thrust/sort.h> 
int main() { 
thrust::device_vector<int> dv; 
thrust::sort(dv.begin(), dv.end()); 
}
`

when compiled with -gpu=cc60 and then run it on a system with cc80, the error message would be:

`
terminate called after throwing an instance of 'thrust::system::system_error' what():  radix_sort: failed on 1st step: cudaErrorUnsupportedPtxVersion: the provided PTX was compiled with an unsupported toolchain. Aborted
`

This doesn't help user to understand what's happening. I tried to address it in this change so that better message will show up:
Incompatible GPU: you are trying to run this program on sm_80, different from the one that it was compiled for.

### DIFF
--- a/thrust/system/cuda/detail/core/util.h
+++ b/thrust/system/cuda/detail/core/util.h
@@ -618,7 +618,8 @@ namespace core {
   inline cuda_optional<int> get_ptx_version()
   {
     int ptx_version = 0;
-    cudaError_t status = cudaGetDevice(&device);
+    int dev_id = 0;
+    cudaError_t status = cudaGetDevice(&dev_id);
     if (status != cudaSuccess)
     {
       throw thrust::system_error(status, thrust::cuda_category(), "No GPU is available\n");
@@ -632,12 +633,12 @@ namespace core {
       int major = 0, minor = 0;
       cudaError_t attr_status;
 
-      attr_status = cudaDeviceGetAttribute(&major, cudaDevAttrComputeCapabilityMajor, device);
+      attr_status = cudaDeviceGetAttribute(&major, cudaDevAttrComputeCapabilityMajor, dev_id);
       cuda_cub::throw_on_error(attr_status,
                               "get_ptx_version :"
                               "failed to get major CUDA device compute capability version.");
 
-      attr_status = cudaDeviceGetAttribute(&minor, cudaDevAttrComputeCapabilityMinor, device);
+      attr_status = cudaDeviceGetAttribute(&minor, cudaDevAttrComputeCapabilityMinor, dev_id);
       cuda_cub::throw_on_error(attr_status,
                               "get_ptx_version :"
                               "failed to get minor CUDA device compute capability version.");
--- a/thrust/system/cuda/detail/core/util.h
+++ b/thrust/system/cuda/detail/core/util.h
@@ -648,7 +648,7 @@ namespace core {
       char str[] = "This program was not compiled for SM     \n";
 
       auto print_1_helper = [&](int v) {
-        str[code_offset] = v + '0';
+        str[code_offset] = static_cast<char>(v) + '0';
         code_offset++;
       };
 
--- a/thrust/system/cuda/detail/core/util.h
+++ b/thrust/system/cuda/detail/core/util.h
@@ -618,16 +618,18 @@ namespace core {
   THRUST_RUNTIME_FUNCTION
   inline int get_ptx_version()
   {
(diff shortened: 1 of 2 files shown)


GPUtester:

dkolsen-pgi:
This is essentially what I was thinking for how to solve this problem.  I don't know if the details are correct or if the code is as efficient as it should be, but having Thrust detect a bad GPU before invoking any kernels and throwing an exception with a good message is the correct approach.

senior-zero (approved):
@zkhatami, @dkolsen-pgi I've made a few adjustments to make it work on both cpu and gpu sides. I've also changed the signature, since the semantic of a function that returns an optional but always throws or returns a value is controversial. Please, take a look and let me know if you agree with the changes.

senior-zero:
run tests

thrust/system/cuda/detail/core/util.h
```diff
@@ -615,11 +616,56 @@ namespace core {
...
+      // Assume two digits will be enough
+      auto print_2_helper = [&](int v) {
+        if (v / 10 != 0) {
+          print_1_helper(v / 10);
+        }
+        print_1_helper(v % 10);
+      };
+
+      print_2_helper(major);
+      print_2_helper(minor);
```
miscco, line 663: I am wondering if it would be simpler to write
```suggestion
      // Assume two digits will be enough
      if (major > 9) {
        str[code_offset++] = (major / 10) + '0';
      }
      str[code_offset++] = major + '0';
        
      if (minor> 9) {
        str[code_offset++] = (minor / 10) + '0';
      }
      str[code_offset++] = minor + '0';
```

thrust/system/cuda/detail/core/util.h
```diff
@@ -615,11 +616,56 @@ namespace core {
   };
 
   THRUST_RUNTIME_FUNCTION
-  inline cuda_optional<int> get_ptx_version()
+  inline int get_ptx_version()
   {
     int ptx_version = 0;
```
miscco, line 621: Nit: we should move this down to the first place of usage

thrust/system/cuda/detail/core/util.h
```diff
@@ -615,11 +616,56 @@ namespace core {
...
-    cudaError_t status = cub::PtxVersion(ptx_version);
-    return cuda_optional<int>(ptx_version, status);
+    const int current_device = cub::CurrentDevice();
+
```
miscco, line 623:
```suggestion
```

miscco (approved):
Looks good to me with a minor request for simplification

senior-zero:
run tests

thrust/system/cuda/detail/core/util.h
```diff
@@ -615,11 +616,55 @@ namespace core {
...
 
   THRUST_RUNTIME_FUNCTION
-  inline cuda_optional<int> get_ptx_version()
+  inline int get_ptx_version()
   {
+    const int current_device = cub::CurrentDevice();
+    if (current_device < 0)
+    {
+      cuda_cub::throw_on_error(cudaErrorNoDevice, "No GPU is available\n");
+    }
```
dkolsen-pgi, line 625: I think this call to `cub::CurrentDevice()` and throwing the exception if it fails should be moved just inside the `if (cub::PtxVersion(ptx_version) != cudaSuccess)` block.  That results in one less CUDA runtime call when we are on the happy path of having the right GPU.  This suggestion assumes that `cub::PtxVersion` will return something other than success when there is no GPU.

senior-zero:
run tests

thrust/system/cuda/detail/core/util.h
senior-zero: I've missed that optimization, thanks!

zkhatami:
Looks good to me as well. Thanks!

senior-zero:
run tests

PR #1855, opened 2023-01-26
Remove deprecated error codes

miscco:
Those error codes are deprecated so we should not reference them any more.

We did not have an internal representation of `cudaErrorApiFailureBase` so I replaced that with `cudaErrorUnknown` which is the numerical predecessor.

Fixes nvbug3896330


miscco:
run tests

senior-zero (approved):

miscco:
run tests

allisonvacanti (approved):

PR #1899, opened 2023-03-22
Add missing #include <exception>

alexfh:
#include <exception> is needed, for example, for std::terminate(). This symbol was likely exported by other standard library headers, but a recent change to LLVM libc++ (https://reviews.llvm.org/D146097) removed some transitive includes of <exception>.


GPUtester:

miscco (approved):

miscco:
Thanks a lot for the PR, I have merged this as is because the change is minimal and obviously correct

PR #1963, opened 2023-06-15
gdb pretty printer: handle non-cuda device vectors

siboehm:
The pretty printer assumes that all device vectors live in CUDA mem space. When using e.g. TBB, this fails with: `$1 = Python Exception <class 'gdb.error'>: No symbol "cudaError" in current context.`.


GPUtester:

miscco:
Hi @siboehm thanks for the contribution.

We have moved our different libraries into a single monorepository. Could you please reopen the PR against https://github.com/NVIDIA/cccl ?

siboehm:
https://github.com/NVIDIA/cccl/pull/264

PR #1986, opened 2024-02-08
:recycle::memo: Update `mode` example to use `thrust::unique_count`

codereport:
When this example was created, `thrust::unique_count` did not exist. It was added in 2022 (2 years ago) in this PR: https://github.com/NVIDIA/thrust/issues/1612. It shouldn't definitely be used now.

I noticed this while @brycelelbach and I were chatting about this algorithm while recording an episode of ADSP 😃

### DIFF
--- a/examples/mode.cu
+++ b/examples/mode.cu
@@ -46,11 +46,7 @@ int main(void)
     std::cout << std::endl;
 
     // count number of unique keys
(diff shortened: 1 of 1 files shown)


codereport:
Closing and moved to CCCL repo: https://github.com/NVIDIA/cccl/pull/1354

